PR #1, opened 2019-07-08
Beginnings of create-single-spa and generator-single-spa

joeldenning:
This project  will use lerna to maintain a monorepo that allows people to create projects with things like `npm init single-spa --framework=react`

### DIFF
--- a/packages/create-single-spa/bin/create-single-spa.js
+++ b/packages/create-single-spa/bin/create-single-spa.js
@@ -1,0 +1,8 @@
+#!/usr/bin/env node
+
+const yeoman = require('yeoman-environment')
+const argv = require('yargs').argv
+
+const env = yeoman.createEnv()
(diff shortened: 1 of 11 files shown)


packages/create-single-spa/bin/create-single-spa.js
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env node
```
joeldenning: This is the file that gets run when the user runs `npm init single-spa`

packages/create-single-spa/bin/create-single-spa.js
```diff
@@ -0,0 +1,8 @@
...
+
+const yeoman = require('yeoman-environment')
```
joeldenning: yeoman will be easier i think to manage than a homegrown CLI thing. A separate npm package called `generator-single-spa` will do the actual generation of all the files and stuff

PR #5, opened 2019-12-14
Creating webpack-config-single-spa-react package.

joeldenning:
This is sort of similar to canopy-webpack-config. Will be used by those who do `npm init single-spa --framework=react`


packages/webpack-config-single-spa-react/lib/webpack-config-single-spa-react.js
```diff
@@ -0,0 +1,104 @@
+const path = require("path");
+const CleanWebpackPlugin = require("clean-webpack-plugin").CleanWebpackPlugin;
+const { BundleAnalyzerPlugin } = require("webpack-bundle-analyzer");
+const { UnusedFilesWebpackPlugin } = require('unused-files-webpack-plugin')
+
+module.exports = webpackConfigSingleSpaReact;
+
+function webpackConfigSingleSpaReact(opts) {
```
frehner: Any desire to use webpack merge or something like that so people can extend it if they want?

packages/webpack-config-single-spa-react/lib/webpack-config-single-spa-react.js
```diff
@@ -0,0 +1,104 @@
...
+      "react-dom",
+      "single-spa",
+      new RegExp(`^@${opts.orgName}/`),
+    ],
+    plugins: [
+      new CleanWebpackPlugin(),
+      new BundleAnalyzerPlugin({
+        analyzerMode: webpackConfigEnv.analyze ? "server" : "disabled"
+      }),
+      new UnusedFilesWebpackPlugin({
```
frehner: I don’t mind this for my code, but it may be a little annoying for people that don’t want it

packages/webpack-config-single-spa-react/lib/webpack-config-single-spa-react.js
```diff
@@ -0,0 +1,104 @@
...
+              loader: "css-loader",
+              options: {
+                modules: false
+              }
+            }
+          ]
+        }
+      ]
+    },
+    devtool: "sourcemap",
```
frehner: Any desire to use a different dev tool in dev mode vs prod?

packages/webpack-config-single-spa-react/lib/webpack-config-single-spa-react.js
joeldenning: We are using webpack-merge - just inside of the user's webpack.config.js file instead of inside of webpack-config-single-spa-react. See

packages/webpack-config-single-spa-react/lib/webpack-config-single-spa-react.js
joeldenning: Why? I always want the best source map possible, regardless of whether I'm in dev or prod mode.

packages/webpack-config-single-spa-react/lib/webpack-config-single-spa-react.js
joeldenning: I view create-single-spa as an opinionated library that doesn't have to please everyone. If people complain about features within it, we can add more `opts` to configure its behavior.

packages/webpack-config-single-spa-react/lib/webpack-config-single-spa-react.js
frehner: ah gotcha

PR #8, opened 2019-12-18
Adding prompts and options for generating root configs and raw in-browser modules

joeldenning:
This sets things up for us to implement a generator for single-spa root configs and also for raw in-browser modules intended for creating styleguides, api caches, etc.

```js
npm init single-spa -- --moduleType ss-app-parcel --framework react
```

### DIFF
--- a/packages/generator-single-spa/src/generator-single-spa.js
+++ b/packages/generator-single-spa/src/generator-single-spa.js
@@ -68,8 +68,8 @@ async function runFrameworkGenerator() {
     case "react":
       this.composeWith({
         Generator: SingleSpaReactGenerator,
-        path: require.resolve('./react/generator-single-spa-react.js')
-      })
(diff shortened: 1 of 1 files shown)


packages/generator-single-spa/src/generator-single-spa.js
```diff
@@ -9,33 +9,72 @@ module.exports = class SingleSpaGenerator extends Generator {
...
+
+    if (!moduleType) {
       const answers = await this.prompt([
         {
-          type: "list",
-          name: "framework",
-          message: "Which framework do you want to use?",
+          type: 'list',
+          name: 'moduleType',
+          message: 'Which kind of in-browser module do you want create-single-spa to create for you?',
```
TheMcMurder: Suggestion: Maybe instead: `Select type to generate`

joeldenning: 👍 updated

PR #9, opened 2019-12-18
Implementing generator for single-spa root configs.

joeldenning:


packages/generator-single-spa/src/generator-single-spa.js
```diff
@@ -28,13 +29,16 @@ module.exports = class SingleSpaGenerator extends Generator {
             { name: 'single-spa root config', value: 'root-config' },
           ]
         }
-      ])
-
-      moduleType = answers.moduleType
+      ])).moduleType
     }
 
+    console.log('moduleType', moduleType)
```
TheMcMurder: I'm guessing we don't want to keep this here long term. Is this intentional or left over from testing?

packages/generator-single-spa/src/root-config/templates/src/index.ejs
```diff
@@ -0,0 +1,30 @@
...
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta http-equiv="X-UA-Compatible" content="ie=edge">
+  <title>Root Config</title>
+  <meta name="importmap-type" content="systemjs-importmap" />
+  <script type="systemjs-importmap" src="https://storage.googleapis.com/react.microfrontends.app/importmap.json"></script>
+  <% if (isLocal) { %>
+  <script type="systemjs-importmap">
+    {
+      "imports": {
+        "@org-name/root-config": "http://localhost:9000/root-config.js"
```
TheMcMurder: is `@org-name` templatized or is that static?

packages/generator-single-spa/src/root-config/templates/src/index.ejs
```diff
@@ -0,0 +1,30 @@
...
+  </script>
+  <% } %>
+  <script src="https://cdn.jsdelivr.net/npm/import-map-overrides/dist/import-map-overrides.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/systemjs/dist/system.min.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/systemjs/dist/extras/amd.min.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/systemjs/dist/extras/named-exports.min.js"></script>
+</head>
+<body>
+  <script>
+    System.import('@org-name/root-config');
```
TheMcMurder: same question about `@org-name` It looks static to me.

packages/generator-single-spa/src/root-config/templates/src/root-config.js
```diff
@@ -0,0 +1,10 @@
+import { registerApplication, start } from "single-spa";
+import * as isActive from "./activity-functions";
+
+registerApplication(
+  "@org-name/navbar",
```
TheMcMurder: Again here

packages/generator-single-spa/src/generator-single-spa.js
joeldenning: Removed this.

packages/generator-single-spa/src/root-config/templates/src/index.ejs
joeldenning: It's just static. Should add a prompt for this, was too lazy to do so earlier

joeldenning:
Updated with the org name being a prompt now and templatized.

PR #10, opened 2019-12-20
Small fixes to react and root config generators

joeldenning:
These were fixes I discovered when running create-single-spa on the react microfrontends example. See related https://github.com/react-microfrontends/nav-bar/pull/2 and https://github.com/react-microfrontends/root-config/pull/1

### DIFF
--- a/packages/generator-single-spa/src/react/templates/package.json
+++ b/packages/generator-single-spa/src/react/templates/package.json
@@ -11,7 +11,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "pretty-quick --staged && concurrently <%= packageManager %>:test <%= packageManager %>:lint <%= packageManager %>:typescript <%= packageManager %>:coverage"
+      "pre-commit": "pretty-quick --staged && concurrently <%= packageManager %>:test <%= packageManager %>:lint"
     }
   },
   "devDependencies": {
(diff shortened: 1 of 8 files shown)


packages/generator-single-spa/src/react/templates/package.json
```diff
@@ -11,7 +11,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "pretty-quick --staged && concurrently <%= packageManager %>:test <%= packageManager %>:lint <%= packageManager %>:typescript <%= packageManager %>:coverage"
```
joeldenning: Copy/paste error with typescript stuff. And no need to run both test and coverage during pre-commit

packages/generator-single-spa/src/root-config/templates/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "scripts": {
-    "start": "webpack-dev-server --mode=development --port 9000 --env.local=true",
+    "start": "webpack-dev-server --mode=development --port 9000 --env.isLocal=true",
```
joeldenning: See https://github.com/CanopyTax/create-single-spa/blob/master/packages/generator-single-spa/src/root-config/templates/webpack.config.js#L31 - was just a typo

PR #12, opened 2020-01-08
Adding support for Vue and Angular CLIs

joeldenning:
This assumes global installation of angular cli or vue cli when you use create-single-spa. But that seems to me to be a reasonable expectation?


packages/generator-single-spa/src/angular/generator-single-spa-angular.js
```diff
@@ -0,0 +1,8 @@
+const Generator = require('yeoman-generator')
+const { spawn } = require('child_process')
+
+module.exports = class SingleSpaAngularGenerator extends Generator {
+  async runVueCli() {
```
EvanBurbidge: nitpick: should this not be runAngularCli?

joeldenning: 👍 updated

PR #13, opened 2020-01-08
Implementing util module creation

joeldenning:
This is for creating utility modules such as a styleguide or api module. They are in-browser modules but are not single-spa applications or parcels.

Usage: (docs to come in the next week or so)

```sh
yarn create single-spa --moduleType=util-module

# Or

npm init single-spa -- --moduleType=util-module
```


packages/webpack-config-single-spa-react/package.json
```diff
@@ -27,9 +27,6 @@
     "url": "https://github.com/CanopyTax/create-single-spa/issues"
   },
   "dependencies": {
-    "clean-webpack-plugin": "^3.0.0",
```
joeldenning: These are dependencies of webpack-config-single-spa, so no need to add them again here

packages/webpack-config-single-spa/package.json
```diff
@@ -17,10 +17,13 @@
...
   },
-  "scripts": {
-    "test": "echo \"Error: run tests from root\" && exit 1"
-  },
+  "scripts": {},
   "bugs": {
     "url": "https://github.com/CanopyTax/create-single-spa/issues"
+  },
+  "dependencies": {
+    "clean-webpack-plugin": "^3.0.0",
```
joeldenning: These dependencies were always needed, but were missing

PR #20, opened 2020-03-04
Adding babel plugin transform runtime. Resolves #17.

### LINKED ISSUE #17: add babel plugin transform runtime
https://babeljs.io/docs/en/babel-plugin-transform-runtime

joeldenning:
See #17. This avoids the `regeneratorRuntime` problems and also saves bundle size

### DIFF
--- a/packages/generator-single-spa/src/react/templates/.babelrc
+++ b/packages/generator-single-spa/src/react/templates/.babelrc
@@ -4,6 +4,9 @@
     "@babel/preset-react"
   ],
   "plugins": [
-    "@babel/plugin-transform-runtime"
+    ["@babel/plugin-transform-runtime", {
+      "useESModules": true,
+      "regenerator": false
+    }]
   ]
 }
--- a/packages/generator-single-spa/src/react/templates/.babelrc
+++ b/packages/generator-single-spa/src/react/templates/.babelrc
@@ -4,5 +4,6 @@
     "@babel/preset-react"
   ],
   "plugins": [
+    "@babel/plugin-transform-runtime"
   ]
 }
--- a/packages/generator-single-spa/src/react/templates/package.json
+++ b/packages/generator-single-spa/src/react/templates/package.json
@@ -19,6 +19,7 @@
     "@babel/plugin-transform-runtime": "^7.8.3",
     "@babel/preset-env": "^7.7.6",
     "@babel/preset-react": "^7.7.4",
+    "@babel/runtime": "^7.0.0-beta.3",
     "@testing-library/react": "^9.4.0",
     "@types/jest": "^24.0.23",
     "babel-eslint": "^11.0.0-beta.2",
--- a/packages/generator-single-spa/src/react/templates/package.json
+++ b/packages/generator-single-spa/src/react/templates/package.json
@@ -19,7 +19,7 @@
     "@babel/plugin-transform-runtime": "^7.8.3",
     "@babel/preset-env": "^7.7.6",
     "@babel/preset-react": "^7.7.4",
-    "@babel/runtime": "^7.0.0-beta.3",
+    "@babel/runtime": "^7.8.7",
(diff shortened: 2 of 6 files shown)


joeldenning:
image

https://babeljs.io/docs/en/babel-plugin-transform-runtime

frehner:
Screen Shot 2020-03-03 at 9 55 39 PM

?

joeldenning:
Hmm yeah maybe need to think about this more. Will look at it tomorrow

frehner:
I mean, I could be totally wrong - like I said, this stuff always manages to confuse me. But at least that's my interpretation of those docs 🤷‍♀

joeldenning:
Interesting commentary on this that I don't quite understand even after having read it a few times - https://github.com/rollup/rollup-plugin-babel#helpers

I created a demo repo that shows that using the runtime helpers results in bundle size savings (search for `TypeError` to see why the transform runtime helps) when you are not using the regeneratorRuntime:

- [Bundle with transform-runtime plugin](https://github.com/joeldenning/webpack-runtime-helpers/blob/with-helpers/dist/prod.js)
- [Bundle without helpers](https://github.com/joeldenning/webpack-runtime-helpers/blob/master/dist/prod.js)

[This documentation](https://babeljs.io/docs/en/next/babel-plugin-transform-runtime.html#technical-details) seems the most clear. I will put comments on my upcoming pushed code with explanations for why I think they are good defaults.

packages/generator-single-spa/src/react/templates/.babelrc
```diff
@@ -4,5 +4,9 @@
     "@babel/preset-react"
   ],
   "plugins": [
+    ["@babel/plugin-transform-runtime", {
+      "useESModules": true,
+      "regenerator": false
```
joeldenning: It's arguably irresponsible to include the regenerator runtime into every bundle that needs to compile async/await or generators to ES5. It is [24kb](https://unpkg.com/browse/regenerator-runtime@0.13.3/runtime.js) that can be repeated in every webpack bundle. With the `regenerator` option turned off, it will assume that `regeneratorRuntime` is a global variable that was only loaded once. For those who don't want to deal with including regenerator runtime into their html file, they can simply change this to true.

packages/generator-single-spa/src/react/templates/.babelrc
```diff
@@ -4,5 +4,9 @@
     "@babel/preset-react"
   ],
   "plugins": [
+    ["@babel/plugin-transform-runtime", {
+      "useESModules": true,
```
joeldenning: This seems to create some bundle savings for webpack bundles - https://babeljs.io/docs/en/next/babel-plugin-transform-runtime.html#useesmodules

packages/generator-single-spa/src/react/templates/.babelrc
```diff
@@ -4,5 +4,9 @@
     "@babel/preset-react"
   ],
   "plugins": [
+    ["@babel/plugin-transform-runtime", {
```
joeldenning: The main bundle savings that this creates is with the [helpers option](https://babeljs.io/docs/en/next/babel-plugin-transform-runtime.html#helpers) defaulting to true - no need to redefine `classCallCheck` multiple times within a single webpack bundle. I omitted that option from this config since it's default is what we want.

packages/generator-single-spa/src/react/templates/package.json
```diff
@@ -16,8 +16,10 @@
   },
   "devDependencies": {
     "@babel/core": "^7.7.5",
+    "@babel/plugin-transform-runtime": "^7.8.3",
     "@babel/preset-env": "^7.7.6",
     "@babel/preset-react": "^7.7.4",
+    "@babel/runtime": "^7.8.7",
```
joeldenning: From docs:

> Make sure you include @babel/runtime as a dependency.

PR #21, opened 2020-03-04
Using js jest config instead of json. Resolves #16

### LINKED ISSUE #16: jest config json not being used
The default path for jest config is `jest.config.js`, not `jest.config.json` (at least as of jest 25). We should update to that. See more at https://jestjs.io/docs/en/configuration#__docusaurus

joeldenning:
See #16


joeldenning:
Oops this PR was incomplete. Luckily the tests caught it. This is ready for re-review

PR #22, opened 2020-03-05
Updated root-config.js so that the template will work out of the box.

kshammer:
Issue: When running npx create-single-spa and going down the root config path the application does not work out of the box as it implies. 

pr2

pr

Solution:
Hard code navbar to be the same as is listed in the given import-map.

### DIFF
--- a/packages/generator-single-spa/src/root-config/generator-root-config.js
+++ b/packages/generator-single-spa/src/root-config/generator-root-config.js
@@ -42,6 +42,7 @@ module.exports = class SingleSpaRootConfigGenerator extends Generator {
         type: "input",
         name: "orgName",
         message: "Organization name (use lowercase and dashes)",
+        default: "react-mf"
       },
     ])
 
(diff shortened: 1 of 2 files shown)


joeldenning:
Hi @kshammer, create-single-spa should prompt you for the organization name, and then fill that in. `@react-mf` is the name of the demo organization, but you'll want to change it to your actual organization's name.

Are you not seeing the prompt for organization name when running create-single-spa?

image

kshammer:
Hi @joeldenning ,

Thank you for getting back to me so fast. I did see the prompt and I left it blank. The reason I made this change is I believe when using these templating tools irregardless of the input the output should work without intervention.  I also believe this will help with the new developer experience. I realize that your team has a done a excellent job of this on the single-spa page with the sandbox and the example applications. 

Thanks,
Kshammer

joeldenning:
> The reason I made this change is I believe when using these templating tools irregardless of the input the output should work without intervention.

I disagree - it is reasonable to expect the user to fill out the prompts. If you'd like to update this PR to provide a default value for the orgName instead of overwriting it to react-mf, I would accept that PR.

> I realize that your team has a done a excellent job of this on the single-spa page with the sandbox and the example applications.

Thanks! I hope you find create-single-spa and single-spa helpful.

packages/generator-single-spa/src/root-config/generator-root-config.js
```diff
@@ -42,35 +42,36 @@ module.exports = class SingleSpaRootConfigGenerator extends Generator {
         type: "input",
         name: "orgName",
         message: "Organization name (use lowercase and dashes)",
+        default: "react-mf"
```
joeldenning: We shouldn't have the organization name default to react-mf, since that's the name of the demo application.

```suggestion
        default: "org-name"
```

Issue #29, opened 2020-03-10
Create application -> vue or angular errors on windows


sullyTheDev:
steps to replicate:

- be on windows
- install latest version globally
- make new dir and cd into it
- ```
    create-single-spa
    ? Select type to generate single-spa application / parcel
    ? Which framework do you want to use? vue
    events.js:187
    throw er; // Unhandled 'error' event
    ^
```

Error: spawn npx ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:240:19)
    at onErrorNT (internal/child_process.js:415:16)
    at process._tickCallback (internal/process/next_tick.js:63:19)
Emitted 'error' event at:
    at Process.ChildProcess._handle.onexit (internal/child_process.js:246:12)
    at onErrorNT (internal/child_process.js:415:16)
    at process._tickCallback (internal/process/next_tick.js:63:19)
error Command failed.
Exit code: 1

I think the cause of this is due to the fact that Node does not support PATHEXT on Windows...so something like this would be a good place to start.

```
    const npxCmd = process.platform === 'win32' ? 'npx.cmd' : 'npx';
    ...
    spawn(npxCmd, ['@angular/cli', 'new'], { stdio: 'inherit' })
    spawn(npxCmd, ['@vue/cli', 'create', '.'], { stdio: 'inherit' })
````

Thanks!

Chris Sullivan

joeldenning:
Hi @sullyTheDev thanks for the clear bug report. Yes, let's fix this to work on windows. Would you able to submit a pull request that adds the windows check? If so, could you please do the same change for the angular generator, so that the same problem does not exist for angular?

joeldenning:
Fixed in #33 and released in https://github.com/single-spa/create-single-spa/releases/tag/v1.3.2 - thanks @sullyTheDev

PR #33, opened 2020-03-11
add a check for windows

sullyTheDev:
added a check for windows to determine to run 'npx' or 'npx.cmd' for ……child process


sullyTheDev:
This fixes issue #29 if someone can/wants to link it 👍

PR #34, opened 2020-03-12
setting up initializion with gitignore

### LINKED ISSUE #32: start with a gitignore
I think we should initialize the created repos with a gitignore file. 

Perhaps something simple like this:

```
# logs
logs
*.log
npm-debug.log*
yarn-debug.log*
yarn-error.log*
# dependencies
node_modules

# OS stuff
.DS_Store
```

TheMcMurder:
Fixes #32 for React

I realized that angular-cli and the vue-cli give us this and the git init for free so this doesn't need to be outside the react generator.


packages/generator-single-spa/src/react/templates/.gitignore
```diff
@@ -0,0 +1,71 @@
...
+dist
+
+# Editor directories and files
+.idea
+.vscode
+*.suo
+*.ntvs*
+*.njsproj
+*.sln
+*.sw?
```
TheMcMurder: This was included in vue-cli's default .gitignore I thought it wouldn't hurt.

packages/generator-single-spa/src/react/templates/.gitignore
```diff
@@ -0,0 +1,71 @@
...
+
+# Yarn Integrity file
+.yarn-integrity
+
+# dotenv environment variables file
+.env
+
+# next.js build output
+.next
+dist
```
TheMcMurder: Everything above this line is the default `.gitignore` for node projects

PR #44, opened 2020-03-28
Add initial CSP meta

### LINKED ISSUE #36: Add CSP to root config html file
See https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP

I think a meta element in the html would be the best way. We need to allow for import map overrides to localhost, but I think other than that should only safe list jsdelivr and unpkg for basically everything (scripts, fetch, etc)

filoxo:
This is an initial pass at adding the CSP header. Resolves #36 

I ran this CSP on https://csp-evaluator.withgoogle.com/ to check for best practices since I'm not very familiar with CSP. 

- Allows unpkg and jsdelivr URLs though this is what the csp-evaluator has to say:
    > cdn.jsdelivr.net is known to host Angular libraries which allow to bypass this CSP.
    
    and the above would also apply to unpkg
- set `object-src` policy to `'none'` as recommended by csp-evaluator
- `script-src` needs `'unsafe-inline'` because (I believe) then the inline importmaps wouldn't work, as well as any inline script tags (at least not without a nonce)
- I have to wonder if `script-src` should also have `'strict-dynamic'` since that would allow a "script to load additional scripts via non-"parser-inserted" script elements (for example document.createElement('script'); is allowed)." which __system.js__ might use.

### DIFF
--- a/packages/generator-single-spa/src/root-config/templates/src/index.ejs
+++ b/packages/generator-single-spa/src/root-config/templates/src/index.ejs
@@ -5,7 +5,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Root Config</title>
-  <meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'unsafe-inline' unpkg.com cdn.jsdelivr.net; object-src 'self';">
+  <meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'unsafe-inline' unpkg.com cdn.jsdelivr.net; object-src 'none';">
   <meta name="importmap-type" content="systemjs-importmap" />
   <script type="systemjs-importmap" src="https://storage.googleapis.com/react.microfrontends.app/importmap.json"></script>
   <% if (isLocal) { %>
--- a/packages/generator-single-spa/src/root-config/templates/src/index.ejs
+++ b/packages/generator-single-spa/src/root-config/templates/src/index.ejs
@@ -5,6 +5,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <meta http-equiv="X-UA-Compatible" content="ie=edge">
   <title>Root Config</title>
+  <meta http-equiv="Content-Security-Policy" content="default-src 'self'; script-src 'unsafe-inline' unpkg.com cdn.jsdelivr.net; object-src 'self';">
   <meta name="importmap-type" content="systemjs-importmap" />
   <script type="systemjs-importmap" src="https://storage.googleapis.com/react.microfrontends.app/importmap.json"></script>
   <% if (isLocal) { %>


joeldenning:
Thanks for your work on this. I think this will be a good improvement to encourage good security practices.

> Allows unpkg and jsdelivr URLs though this is what the csp-evaluator has to say:

Localhost is also needed for import map overrides.

> script-src needs 'unsafe-inline' to allow any inline script tags and therefore (I believe) inline importmaps wouldn't work

Inline import maps aren’t executed by the browser at all with the systemjs setup - systemjs finds the script and parses its text content as json. I think this could be removed.

> I have to wonder if script-src should also have 'strict-dynamic' to allow a "script to load additional scripts via non-"parser-inserted" script elements (for example document.createElement('script'); is allowed)." which system.js might be doing.

Systemjs is definitely inserting dynamic script elements - one for every system.import.

>  I expect this might need more tweaking to get right.

I think the best way to test this would be to add it to vue-microfrontends, polyglot-microfrontends, and react-microfrontends. They each have a root config repo that was generate with create single spa, but don’t have a meta tag. If it works locally and when deployed, including with import map overrides, then that’s the beat sign it’ll be okay here.

joeldenning:
@filoxo are you still planning on working on this?

filoxo:
Yea, just haven't had time as I'm winding down at my current company and going to be starting a new job soon. Feel free to pick it up if I'm not able to get to it as quickly as you need.

filoxo:
Ok, I was able to put this into the vue-microfrontends root config and learned a few things.

> Inline import maps aren’t executed by the browser at all with the systemjs setup - systemjs finds the script and parses its text content as json. I think this could be removed.

It appears that even if it the inline import map doesn't evaluate to anything, the browser blocks it from even being parsed because its inside of a script tag. So it appears this has to stay. 

Here is the CSP that I used that was able to get all the security errors resolved;

```html
<meta http-equiv="Content-Security-Policy" content="default-src 'self' localhost:*; script-src 'unsafe-inline' localhost:* unpkg.com cdn.jsdelivr.net storage.googleapis.com; connect-src localhost:* ws://localhost:* https://dog.ceo https://storage.googleapis.com; object-src 'none'; style-src 'unsafe-inline' https://fonts.googleapis.com; font-src https://fonts.gstatic.com; img-src localhost:* https://images.dog.ceo https://external-content.duckduckgo.com;">
```

I'm going to try to break down why each part is necessary:

- `default-src 'self' localhost:*;`: this fallback allows requests to localhost and other localhost servers with different ports, for some directives if they are not present. This is probably optional but might be worth leaving in? 
-  `script-src 'unsafe-inline' localhost:* unpkg.com cdn.jsdelivr.net storage.googleapis.com;`
    - `'unsafe-inline'` is needed for inline `<script>` tags, which the import map is when running locally. Without this, the browser throws and error even just trying to parse it (as mentioned above).
    - the remaining URLs are CDNs where external script assets are hosted
- `connect-src localhost:* ws://localhost:* https://*;`
    - these URLs are allowed to be loaded using script interfaces` (XHR, fetch, event source, websockets, and a couple others that aren't commonly used)
- `object-src 'none';` is recommended by Google CSP evaluator 
- `style-src 'unsafe-inline' https://fonts.googleapis.com;`
    - `'unsafe-inline'` is needed because one or more of the MFs dynamically inject style tags
    -  Google font is the external CSS import in [styleguide's global.css](https://github.com/vue-microfrontends/styleguide/blob/master/src/global.css#L1)
- `font-src https://fonts.gstatic.com;` allows loading the font itself from this domain
- `img-src localhost:* https://images.dog.ceo https://external-content.duckduckgo.com;`
    - `localhost:*` are images from MFs
    - `https://images.dog.ceo` are images from dog.ceo api
    - `https://external-content.duckduckgo.com` is the host from where the puppy in the "logo" is loaded from

This makes me worry that it will be a point of frustration for people who don't know or understand CSP. I know I was/am one of them. Or might just need enough documentation around it.

joeldenning:
Nice work putting all this together. I agree that if we can shield our users from having to learn all the nitty gritty of CSP the first time they set up single-spa that that would be great.

One option, perhaps, would be put in a CSP meta tag into the template, but comment it out? And then add a comment above it explaining that it's good for security to enable CSP, but also might cause some pain.

Idk what do you think? Worth it to turn CSP on by default? Or too much of a hurdle for people just getting started?

filoxo:
Yea I would agree with that. What do you think of just enabling all those rules but with generic schemas (eg. `https://`) instead? I'll try that and see how it works. Worst case is we do like you suggested and add it but commented out.

joeldenning:
Generic schemas could be good - obviously less secure, but turning off object-src would be good, and if we add comments explaining what people should do to make it more secure, it could be hlpeful

PR #55, opened 2020-04-06
exploring adding an explanitory comment to the file

TheMcMurder:
I've seen a few questions about asking why the CLI generates this file

https://single-spa.slack.com/archives/C8R6U7MT7/p1586204715291500

and a few I've had internally at workfront. I just wanted to discuss if we should add _something like_ this to help answer some of those questions

### DIFF
--- a/packages/generator-single-spa/src/react/templates/src/set-public-path.js
+++ b/packages/generator-single-spa/src/react/templates/src/set-public-path.js
@@ -5,4 +5,4 @@ import { setPublicPath } from "systemjs-webpack-interop";
  * https://single-spa.js.org/docs/faq/#code-splits
  */
 
-setPublicPath('@<%= orgName %>/<%= projectName %>');
+setPublicPath("@<%= orgName %>/<%= projectName %>");
--- a/packages/generator-single-spa/src/react/templates/src/set-public-path.js
+++ b/packages/generator-single-spa/src/react/templates/src/set-public-path.js
@@ -1,3 +1,8 @@
 import { setPublicPath } from "systemjs-webpack-interop";
+/* This dynamically sets the webpack public path so that code splits work properly. See related:
+ * https://github.com/joeldenning/systemjs-webpack-interop#what-is-this
+ * https://webpack.js.org/guides/public-path/#on-the-fly
+ * https://single-spa.js.org/docs/faq/#code-splits
+ */
 
 setPublicPath('@<%= orgName %>/<%= projectName %>');
--- a/packages/generator-single-spa/src/util-module/templates/src/set-public-path.js
+++ b/packages/generator-single-spa/src/util-module/templates/src/set-public-path.js
@@ -1,17 +1,8 @@
 import { setPublicPath } from "systemjs-webpack-interop";
+/* This dynamically sets the webpack public path so that code splits work properly. See related:
+ * https://github.com/joeldenning/systemjs-webpack-interop#what-is-this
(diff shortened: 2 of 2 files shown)


packages/generator-single-spa/src/util-module/templates/src/set-public-path.js
```diff
@@ -1,3 +1,17 @@
 import { setPublicPath } from "systemjs-webpack-interop";
 
-setPublicPath('@<%= orgName %>/<%= projectName %>');
\ No newline at end of file
+setPublicPath('@<%= orgName %>/<%= projectName %>');
+// This file exists to solve a problem with how webpack loads code splits
```
joeldenning, line 4: I found it a bit weird that this comment was below the code instead of above it.

joeldenning: Also, prettier might be able to format this file really nicely, especially if using `/*`. It will force certain column width.

`npx prettier set-public-path.js --write`

joeldenning: I think this explanation is a bit unclear. "Url origin" could mean a few things. We already have a couple explanations of this at https://single-spa.js.org/docs/faq/#code-splits and https://github.com/joeldenning/systemjs-webpack-interop#what-is-this.

I think I'd prefer just linking to those and then updating those to be good, instead of explaining it here.

```js
/* This dynamically sets the webpack public path so that code splits work properly. See related:
* https://github.com/joeldenning/systemjs-webpack-interop#what-is-this
* https://webpack.js.org/guides/public-path/#on-the-fly
* https://single-spa.js.org/docs/faq/#code-splits
*/
setPublicPath('@<%= orgName %>/<%= projectName %>'); 
```

TheMcMurder: Linking is 1000% better. I'll change this to match.

packages/generator-single-spa/src/util-module/templates/src/set-public-path.js
```diff
@@ -1,3 +1,8 @@
 import { setPublicPath } from "systemjs-webpack-interop";
+/* This dynamically sets the webpack public path so that code splits work properly. See related:
+ * https://github.com/joeldenning/systemjs-webpack-interop#what-is-this
+ * https://webpack.js.org/guides/public-path/#on-the-fly
+ * https://single-spa.js.org/docs/faq/#code-splits
+ */
 
-setPublicPath('@<%= orgName %>/<%= projectName %>');
\ No newline at end of file
+setPublicPath("@<%= orgName %>/<%= projectName %>");
```
TheMcMurder, line 8: This change from prettier

packages/generator-single-spa/src/react/templates/src/set-public-path.js
```diff
@@ -1,3 +1,8 @@
 import { setPublicPath } from "systemjs-webpack-interop";
+/* This dynamically sets the webpack public path so that code splits work properly. See related:
+ * https://github.com/joeldenning/systemjs-webpack-interop#what-is-this
+ * https://webpack.js.org/guides/public-path/#on-the-fly
+ * https://single-spa.js.org/docs/faq/#code-splits
+ */
 
-setPublicPath('@<%= orgName %>/<%= projectName %>');
\ No newline at end of file
+setPublicPath("@<%= orgName %>/<%= projectName %>");
```
TheMcMurder, line 8: This change from prettier

PR #58, opened 2020-04-07
Format with prettier

### LINKED ISSUE #52: Add prettier, husky, pretty-quick, .prettierignore, and `format` script to root package.json
To auto format all files in the create-single-spa repo.

filoxo:
Resolves #52 

Added `prettier` with a precommit hook. To test this, I made a change to the main.js file to destructure and export the lifecycle functions. I feel that this is a more concise syntax and a very high percentage of projects don't need more than this (eg. exporting an array of fns). So my remaining questions are:

- Is the style change ok? I'm not opposed to reverting it, it at least shows some of the potential formatting changes.
- Are the default prettier formatting options good? Or are there other options we'd prefer?


package.json
```diff
@@ -2,11 +2,20 @@
...
+    "lerna": "^3.15.0",
+    "prettier": "^2.0.4",
+    "pretty-quick": "^2.0.1"
   },
   "scripts": {
     "bootstrap": "lerna bootstrap",
     "test": "lerna run test",
-    "lint": "lerna run lint"
+    "lint": "lerna run lint",
+    "format": "pretty-quick"
```
joeldenning, line 14: pretty-quick only runs prettier on changed files - I'd prefer to run the format command to run on *all* files.

```suggestion
    "format": "prettier './**/*' --write"
```

Any files that prettier doesn't have a parser for should be added to a `.prettierignore` file

filoxo: I didn't want to be overly aggressive but this sounds good to me.

PR #51, opened 2020-04-06
Typescript support

brandones:
Fixes https://github.com/single-spa/create-single-spa/issues/38

PR #62, opened 2020-04-08
Improving typescript support. Resolves #61.

### LINKED ISSUE #61: When choosing typescript, npm/yarn install occurs twice
image

See https://github.com/single-spa/create-single-spa/blob/master/packages/generator-single-spa/src/react/generator-single-spa-react.js#L159 and https://github.com/single-spa/create-single-spa/blob/master/packages/generator-single-spa/src/react/generator-single-spa-react.js#L163. Also see https://github.com/single-spa/create-single-spa/pull/51#discussion_r404885125

cc @brandones

joeldenning:
See #61 - this fixes that and also a few other issues with typescript support.


.prettierignore
```diff
@@ -2,3 +2,4 @@
 .prettierignore
 yarn.lock
 *.ejs
+packages/generator-single-spa/src/react/templates/webpack.config.js
```
joeldenning, line 5: This now has ejs templating in it that breaks prettier

packages/generator-single-spa/src/react/generator-single-spa-react.js
```diff
@@ -57,6 +43,35 @@ module.exports = class SingleSpaReactGenerator extends Generator {
...
+    const packageJsonStr = ejs.render(packageJsonTemplate, {
+      packageManager: this.packageManager,
+      typescript: this.typescript,
+    });
+
+    const packageJson = JSON.parse(packageJsonStr);
+
+    if (this.typescript) {
+      // Will be added as a dependency via ts-package.json
+      delete packageJson.devDependencies["@types/jest"];
```
joeldenning, line 60: Previously, `@types/jest` was being added as both a dependency and a devDependency

packages/generator-single-spa/src/react/generator-single-spa-react.js
```diff
@@ -57,6 +43,35 @@ module.exports = class SingleSpaReactGenerator extends Generator {
...
+      // Will be replaced by eslint-config-ts-react-important-stuff
+      delete packageJson.devDependencies["eslint-config-react-important-stuff"];
+    }
+
+    this.fs.extendJSON(this.destinationPath("package.json"), packageJson);
+
+    if (this.typescript) {
+      this.fs.extendJSON(
```
joeldenning, line 68: I switched to using extendJSON and an external ts-package.json file for this, instead of inlining the dependencies within this file. The reason is that I couldn't avoid the double yarn install any other way

packages/generator-single-spa/src/react/generator-single-spa-react.js
```diff
@@ -105,18 +122,18 @@ module.exports = class SingleSpaReactGenerator extends Generator {
     );
     this.fs.copyTpl(
       this.templatePath("src/root.component.js"),
-      this.destinationPath("src/root.component.js"),
+      this.destinationPath(`src/root.component.${srcFileExtension}`),
```
joeldenning, line 125: typescript react files should have `.tsx` file extensions

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -88,8 +96,11 @@ function webpackConfigSingleSpa(opts) {
           ],
         },
       }),
-    ].concat(
-      opts.typecheck === "typescript" ? [new ForkTsCheckerWebpackPlugin()] : []
-    ),
+    ].concat(typescript ? [new ForkTsCheckerWebpackPlugin()] : []),
+    resolve: {
+      extensions: [".js", ".mjs", ".wasm", ".json"].concat(
+        typescript ? [".ts", ".tsx"] : []
```
joeldenning, line 102: So that `import './component'` works for typescript projects without the file extension

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -21,12 +21,20 @@ function webpackConfigSingleSpa(opts) {
...
+    entryFileExtension = opts.react ? "tsx" : "ts";
+  } else {
+    entryFileExtension = "js";
+  }
 
   return {
     entry: path.resolve(
       process.cwd(),
-      `src/${opts.orgName}-${opts.projectName}.js`
+      `src/${opts.orgName}-${opts.projectName}.${entryFileExtension}`
```
joeldenning, line 37: Change file extension based on whether typescript / react are being used

joeldenning:
I tested this out fully by creating a project with this modified code and running `yarn build`, `yarn start`, `yarn prettier`, and `yarn lint` on it. I did so with both a typescript project and a js project, and everything is working.

cc @brandones this is an extension of your PR #51

packages/generator-single-spa/src/react/generator-single-spa-react.js
brandones: Is this better than just inlining the TS dependencies in `package.json.ejs`, now that the latter is templated over `typescript` anyway?

brandones: You could also take care of `      delete packageJson.devDependencies["eslint-config-react-important-stuff"];` that way

packages/generator-single-spa/src/react/generator-single-spa-react.js
TheMcMurder: Dumb question is that required or just community norms?
`.ts` = .typescript
`.jsx` for file contains jsx
`.tsx` for file is typescript and contains jsx.

IIRC It's not a hard requirement to use `.jsx` with webpack is it a hard requirement to use `.tsx` with typescript compiler?

packages/generator-single-spa/src/react/generator-single-spa-react.js
brandones: Not sure I understand. Shouldn't this only be installed if TS is present, and only as a normal dependency? Why is it in `devDependencies` in the first place?

Unless, is the idea that this generator should update previously-generated projects?

packages/generator-single-spa/src/react/templates/package.json
```diff
@@ -3,7 +3,7 @@
     "start": "webpack-dev-server",
     "build": "webpack --mode=production",
     "analyze": "webpack --mode=production --env.analyze=true",
-    "lint": "eslint src",
+    "lint": "eslint src --ext js<% if (typescript) { %>,ts,tsx<% } %>",
```
TheMcMurder, line 6: Dumb queston: Any reason why we couldn't just have all three here always? It wouldn't find `ts` or `tsx` unless you created them.

packages/generator-single-spa/src/react/generator-single-spa-react.js
```diff
@@ -73,6 +88,8 @@ module.exports = class SingleSpaReactGenerator extends Generator {
     this.orgName = templateOptions.orgName;
     this.projectName = templateOptions.projectName;
 
+    const srcFileExtension = this.typescript ? "tsx" : "js";
```
brandones, line 91: Nice, this is way better than templating the filename itself (what I'd done previously for another generator)

packages/webpack-config-single-spa-react/lib/webpack-config-single-spa-react.js
```diff
@@ -3,6 +3,7 @@ const webpackConfigSingleSpa = require("webpack-config-single-spa");
 module.exports = webpackConfigSingleSpaReact;
 
 function webpackConfigSingleSpaReact(opts) {
+  opts.react = true;
```
brandones, line 6: Just curious, what's this do?

packages/generator-single-spa/src/react/generator-single-spa-react.js
joeldenning: It is there for non typescript projects simply so VS Code doesn't freak out or try incorrect autocompletions for `describe()` and `it()`. Non typescript projects can benefit from typescript simply by having their IDE look inside of their node_modules

packages/generator-single-spa/src/react/generator-single-spa-react.js
joeldenning: Yeah I thought of doing it that way, and it would work. But then we couldn't run `yarn upgrade-interactive-latest --interactive` on the template package json to upgrade its dependencies. Which is pretty convenient for keeping things up-to-date. I decided with this approach for that reason, but agree this implementation is maybe less straightforward.

packages/generator-single-spa/src/react/generator-single-spa-react.js
joeldenning: It is a hard requirement of typescript. You absolutely MUST use `.ts` for typescript files, and MUST use `.tsx` if it has JSX in it.

packages/generator-single-spa/src/react/templates/package.json
joeldenning: Responded in https://github.com/single-spa/create-single-spa/pull/62#discussion_r405682730

packages/webpack-config-single-spa-react/lib/webpack-config-single-spa-react.js
joeldenning: https://github.com/single-spa/create-single-spa/pull/62/files#diff-9629c0751be0b0a313fe13ae8b46351eR29

packages/generator-single-spa/src/react/generator-single-spa-react.js
TheMcMurder: I had no idea it was a hard requirement. Good to know. Thank you.

PR #64, opened 2020-04-08
adding missing dependency

TheMcMurder:
This fixes the following error when trying to use `webpack-config-single-spa` (running yarn start) on a newly created application

Screen Shot 2020-04-08 at 4 41 50 PM

### DIFF
--- a/packages/webpack-config-single-spa/package.json
+++ b/packages/webpack-config-single-spa/package.json
@@ -23,6 +23,7 @@
   },
   "dependencies": {
     "clean-webpack-plugin": "^3.0.0",
+    "fork-ts-checker-webpack-plugin": "^4.1.3",
     "unused-files-webpack-plugin": "^3.4.0",
     "webpack-bundle-analyzer": "^3.6.0"
   }
--- a/packages/webpack-config-single-spa/yarn.lock
+++ b/packages/webpack-config-single-spa/yarn.lock
@@ -87,12 +87,37 @@ acorn@^6.0.7:
   resolved "https://registry.yarnpkg.com/acorn/-/acorn-6.4.1.tgz#531e58ba3f51b9dacb9a6646ca4debf5b14ca474"
   integrity sha512-ZVA9k326Nwrj3Cj9jlh3wGFutC2ZornPNARZwsNYqQYgN0EsV2d53w5RN/co65Ohn4sUAUtb1rSUAOD6XN9idA==
 
+ansi-regex@^2.0.0:
+  version "2.1.1"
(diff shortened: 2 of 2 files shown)


packages/webpack-config-single-spa/package.json
```diff
@@ -23,6 +23,7 @@
   },
   "dependencies": {
     "clean-webpack-plugin": "^3.0.0",
+    "fork-ts-checker-webpack-plugin": "^4.1.3",
```
TheMcMurder, line 26: This is used here: https://github.com/single-spa/create-single-spa/blob/master/packages/webpack-config-single-spa/lib/webpack-config-single-spa.js#L5

But it's not in the package.json so it isn't bundled into the npm module and the npm module cannot run without it.

joeldenning: If we add this, we also will need to add `typescript` as a dependency.

Hmm I'd hate to force everyone to install fork-ts-checker-webpack-plugin and typescript even when they're not using them. One option could be to mark them as optionalDependencies?? To hint at the need for them to be installed. Then we'd update the webpack config to only call `require("fork-ts-checker-webpack-plugin")` inside of the ternary so that it's avoided when typescript is not needed?

Alternatively we just bite the bullet and add both of them as dependencies. Or even split this package into two, so that we can have a separate package.json for typescript dependencies. Thoughts?

joeldenning: I honestly think webpack-config-single-spa-ts might be best?

TheMcMurder: This is probably naive but could we just do a dynamic require here instead?

https://github.com/single-spa/create-single-spa/blob/master/packages/webpack-config-single-spa/lib/webpack-config-single-spa.js#L99

I'm not sure I understand why typescript is a dependency of that plugin

TheMcMurder: Or could we have a `webpack-config-single-spa-ts` that just uses webpack merge to add those options to `webpack-config-single-spa`?

joeldenning: Yes that's what I had in mind. webpack-config-single-spa-ts would use webpack-config-single-spa.

joeldenning: > This is probably naive but could we just do a dynamic require here instead?

Yes that is exactly what I suggested in my comment:

> Then we'd update the webpack config to only call require("fork-ts-checker-webpack-plugin") inside of the ternary so that it's avoided when typescript is not needed

TheMcMurder: > Yes that is exactly what I suggested in my comment
🤦‍♂  I read that wrong. Whoops.

As far as dynamic require vs `webpack-config-single-spa-ts` I'm good with either option. What's your preference? This MR won't solve it so I'll close it.

PR #65, opened 2020-04-08
moving gitignore to a common templates directory where it can be used…

TheMcMurder:
… by react and root config projects


joeldenning:
While we're at it, maybe add this gitignore to the util module generator, too?

https://github.com/single-spa/create-single-spa/blob/master/packages/generator-single-spa/src/util-module/generator-single-spa-util-module.js

Issue #70, opened 2020-04-09
Failed to compile because generator does not produce a src directory


protoEvangelion:
The generator is failing to produce applications that compile.  The issue looks like that there is no `/lib/node_modules/create-single-spa/node_modules` folder. 

![image](https://user-images.githubusercontent.com/20076677/78903624-430a7880-7a30-11ea-9487-e706a61fe724.png)

Maybe this has to do with the fact that I'm using [n](https://github.com/tj/n) on a mac to manage node versions?

Here is the output from running the generator:

```
 ~/dev/monorepo| => create-single-spa
? Select type to generate single-spa application / parcel
? Which framework do you want to use? react
? Which package manager do you want to use? yarn
? Will this project use Typescript? No
? Organization name (use lowercase and dashes) tio
? Project name (use lowercase and dashes) lumin
(node:14692) UnhandledPromiseRejectionWarning: AssertionError [ERR_ASSERTION]: Trying to copy from a source that does not exist: /Users/rgarant/n/lib/node_modules/create-single-spa/node_modules/generator-single-spa/src/common-templates/.gitignore
    at EditionInterface.exports.copy (/Users/rgarant/n/lib/node_modules/create-single-spa/node_modules/mem-fs-editor/lib/actions/copy.js:49:3)
    at EditionInterface.module.exports [as copyTpl] (/Users/rgarant/n/lib/node_modules/create-single-spa/node_modules/mem-fs-editor/lib/actions/copy-tpl.js:29:8)
    at SingleSpaReactGenerator.copyOtherFiles (/Users/rgarant/n/lib/node_modules/create-single-spa/node_modules/generator-single-spa/src/react/generator-single-spa-react.js:108:13)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:14692) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:14692) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
   create package.json
   create jest.config.js
   create .babelrc
   create .eslintrc
```

Yarn succeeds & then I run the start cmd: `yarn start --https --port 8500`

```
ERROR in Entry module not found: Error: Can't resolve './src' in '/Users/rgarant/dev/monorepo'

[...]

Issue #41, opened 2020-03-26
Error copying missing .gitignore in generator-single-spa


andreaskc:
```
npm init single-spa
? Select type to generate single-spa application / parcel
? Which framework do you want to use? react
? Which package manager do you want to use? npm
? Organization name (use lowercase and dashes) ls
? Project name (use lowercase and dashes) appbar
events.js:186
   throw er; // Unhandled 'error' event
   ^
AssertionError [ERR_ASSERTION]: Trying to copy from a source that does not exist: /Users/andrewkallenbach/.npm/_npx/99955/lib/node_modules/create-single-spa/node_modules/generator-single-spa/src/react/templates/.gitignore
```

I've tried this on a Mac 10.14 and also a Windows 10 machine (and the Windows Ubuntu subsystem) with same result.
Create-single-spa is v 1.3.3
Node is 12.10

joeldenning:
> Also, should we migrate this issue to the create-single-spa repo?

Yes, done.

> I've reproduced this locally

That's great to hear, I am unable to reproduce it for some reason.

> It appears that npm publish renames .gitignore to .npmignore

Good catch! Why then does unpkg show the gitignore inside of https://unpkg.com/browse/generator-single-spa@1.3.3/src/react/templates/.gitignore ?? The published code appears to have the gitignore - is it being removed during install??? Or is unpkg showing something that's not in the published tarball?

joeldenning:
Note that the gitignores also show up in jsdelivr - https://cdn.jsdelivr.net/npm/generator-single-spa@1.3.3/src/react/templates/.gitignore

filoxo:
Huh I wonder if its converted over while unpacking...

TheMcMurder:
> Huh I wonder if its converted over while unpacking...

After investigation before submitting #71 @filoxo discovered that it's changed by NPM prior to publishing the module.

If I ran `npm pack` it would still be `.gitignore` but once you installed it globally (or used npx) from npm and looked at the code it would be `.npmignore`

PR #71, opened 2020-04-09
.gitignore files are turned into .npmignore files :facepalm: this sho…

TheMcMurder:
…uld fix it #CarlosAlreadyFixedItButIBrokeIt

This should fix issue #70


packages/generator-single-spa/src/react/generator-single-spa-react.js
```diff
@@ -106,7 +106,7 @@ module.exports = class SingleSpaReactGenerator extends Generator {
       templateOptions
     );
     this.fs.copyTpl(
-      this.templatePath("../../common-templates/.gitignore"), // this is relative to /templates
+      this.templatePath("../../common-templates/gitignore"), // this is relative to /templates
```
TheMcMurder, line 109: This is strange. 

NPM reads all .gitignore files and turns them into a .npmignore so in the packaged code there is no .gitignore but if we change this to `gitignore` without the `.` then NPM won't mess with it.

filoxo:
For documentation purposes in case this happens again in the future, here is the original issue: https://github.com/single-spa/create-single-spa/issues/41#issuecomment-604839777

PR #73, opened 2020-04-10
Rename format script

### LINKED ISSUE #40: prettierignore and format script for root config

filoxo:
Resolves #40


filoxo:
Good catch! I updated those as well.

PR #74, opened 2020-04-10
Add jest-dom to React template

### LINKED ISSUE #53: Add `@testing-library/jest-dom` to react template
This is what CRA uses - it allows for assertions such as `expect(getByText('some text')).toBeInDocument()`

filoxo:
Resolves #53 

- Added to React template package.json
- Configured in React template jest.config.js
- Added a very simple test file for root.component

### DIFF
--- a/packages/generator-single-spa/src/react/generator-single-spa-react.js
+++ b/packages/generator-single-spa/src/react/generator-single-spa-react.js
@@ -126,6 +126,11 @@ module.exports = class SingleSpaReactGenerator extends Generator {
       templateOptions
     );
     this.fs.copyTpl(
+      this.templatePath("src/root.component.test.js"),
+      this.destinationPath(`src/root.component.test.${srcFileExtension}`),
+      templateOptions
+    );
+    this.fs.copyTpl(
       this.templatePath("src/set-public-path.js"),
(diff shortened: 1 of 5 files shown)


packages/generator-single-spa/src/react/generator-single-spa-react.js
```diff
@@ -125,6 +125,11 @@ module.exports = class SingleSpaReactGenerator extends Generator {
       this.destinationPath(`src/root.component.${srcFileExtension}`),
       templateOptions
     );
+    this.fs.copyTpl(
```
filoxo, line 128: I'm not sure if this plays well with Typescript. I'm testing this now.

joeldenning:
Nice work. One thing - let’s remove the —passWithNoTests flag to the test command in the package.json, now that there is a test

packages/generator-single-spa/src/react/generator-single-spa-react.js
filoxo: This works just fine with Typescript, but this testing uncovered another issue. The project would not start with TypeScript support because of a plugin constructor invocation problem. That fix is also included in this PR.

PR #76, opened 2020-04-10
Customize destination directory and use global vue/angular when possible. Resolves #56

### LINKED ISSUE #56: Vue and Angular create a new directory but React doesn't
More of a consistency issue than an actual bug.

Right now the CLI leverages vue-cli and angular-cli. Both of those CLIs create a new folder for you. 

Our react option does not.

Do we want to create a directory to be consistent across the possible choices?

If we do then I'll submit a PR

joeldenning:
See #56. Also this substantially speeds things up for angular/vue users who already have global installations of their respective CLIs.

I tested all of the following in all frameworks
```sh
# current dir
create-single-spa
create-single-spa .

# In new dir
create-single-spa app1

# Alternative way of specifying dir
create-single-spa --dir app1
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ We're using [lerna](https://lerna.js.org/).
 2. `yarn bootstrap`
 3. `node packages/create-single-spa/bin/create-single-spa.js ../some-test-dir`
 
-Note that create-single-spa can runs in the current directory by default, but you can provide a different destination directory through the CLI (as shown above with `some-test-dir`).
+Note that create-single-spa can run in the current directory by default, but you can provide a different destination directory through the CLI (as shown above with `some-test-dir`).
 
 ### Tests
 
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -12,9 +12,9 @@ We're using [lerna](https://lerna.js.org/).
 
 1. `yarn install`
 2. `yarn bootstrap`
-3. `node packages/create-single-spa/bin/create-single-spa.js`
+3. `node packages/create-single-spa/bin/create-single-spa.js ../some-test-dir`
 
-Note that create-single-spa **runs in the current directory**. For development purposes, you may want to create a test directory outside of the create-single-spa folder to run it on. There is no compilation / build step.
+Note that create-single-spa can runs in the current directory by default, but you can provide a different destination directory through the CLI (as shown above with `some-test-dir`).
```
filoxo, line 17: typo: `can runs` has an extraneous "s"

joeldenning: 👍

Issue #100, opened 2020-04-21
ERR_INVALID_OPT_VALUE when using create-single-spa


dcalise:
```
$ create-single-spa
(node:59293) ExperimentalWarning: The fs.promises API is experimental
? Directory for new project .
? Select type to generate single-spa root config
? Which package manager do you want to use? npm
? Will this project use Typescript? No
events.js:167
      throw er; // Unhandled 'error' event
      ^

TypeError [ERR_INVALID_OPT_VALUE]: The value "undefined" is invalid for option "flags"
    at stringToFlags (internal/fs/utils.js:253:9)
    at open (internal/fs/promises.js:198:34)
    at Object.readFile (internal/fs/promises.js:466:20)
    at SingleSpaRootConfigGenerator.createPackageJson (/Users/dcalise/.nvm/versions/node/v10.2.1/lib/node_modules/create-single-spa/node_modules/generator-single-spa/src/root-config/generator-root-config.js:49:42)
    at process._tickCallback (internal/process/next_tick.js:68:7)
Emitted 'error' event at:
    at Immediate.setImmediate (/Users/dcalise/.nvm/versions/node/v10.2.1/lib/node_modules/create-single-spa/node_modules/yeoman-generator/lib/index.js:791:20)
    at runCallback (timers.js:696:18)
    at tryOnImmediate (timers.js:667:5)
    at processImmediate (timers.js:649:5)
```

I've tried both via `npx create-single-spa` as well is an `npm -g` install and `create-single-spa` command.

joeldenning:
Please upgrade your version of nodejs to 10.4.0 or later. I see you're on 10.2.1, which had a bug that was fixed that create-single-spa relies on.

I'm closing this issue and will add the minimum nodejs requirements to the documentation.

joeldenning:
I decided to add a check to create-single-spa instead of just adding it to the documentation. #101 adds a minimum nodejs check to the start of create-single-spa so that this will be surfaced better for those in the future

joeldenning:
The check for NodeJS version is implemented in 1.7.0

PR #101, opened 2020-04-21
Adding check for minimum nodejs version

joeldenning:
See related https://github.com/single-spa/create-single-spa/issues/100

### DIFF
--- a/packages/create-single-spa/bin/create-single-spa.js
+++ b/packages/create-single-spa/bin/create-single-spa.js
@@ -31,6 +31,12 @@ function checkNodeVersion() {
   }
 
   if (!validVersion) {
-    throw Error("create-single-spa requires NodeJS >= " + minVersion);
+    throw Error(
+      "create-single-spa requires NodeJS >= " +
+        minVersion +
+        ", but you are using NodeJS " +
+        process.versions.node +
+        "."
+    );
   }
 }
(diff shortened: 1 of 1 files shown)


packages/create-single-spa/bin/create-single-spa.js
```diff
@@ -1,8 +1,36 @@
 #!/usr/bin/env node
 
+checkNodeVersion();
+
 const yeoman = require("yeoman-environment");
 const argv = require("yargs").argv;
 
 const env = yeoman.createEnv();
 env.registerStub(require("generator-single-spa"), "single-spa");
-env.run(`single-spa ${argv._.join(" ")}`, argv);
```
joeldenning, line 8: who knows - maybe someone will be using a version of node that doesn't support template strings. Poor souls

packages/create-single-spa/bin/create-single-spa.js
```diff
@@ -1,8 +1,36 @@
...
+  } else if (currentMajor === minMajor) {
+    if (currentMinor < minMinor) {
+      validVersion = false;
+    } else if (currentMinor === minMinor && currentPatch < minPatch) {
+      validVersion = false;
+    }
+  }
+
+  if (!validVersion) {
+    throw Error("create-single-spa requires NodeJS >= " + minVersion);
```
filoxo, line 34: I find it very helpful when the tool also shows what it received vs what is expected. What do you think of adding the detected version of node to this error message?

TheMcMurder: Great idea!

joeldenning: updated

PR #124, opened 2020-05-20
Add @single-spa/welcome application

filoxo:
This attempts to create a welcome application that is imported and shown when creating a root config with create-single-spa. This application hopefully guides users with steps on what to do next. I'm not super confident that the steps I've included are the best for learning and getting started, but something is probably better than nothing :) Here is a screenshot of the application running on my own local root-config I generated with c-s-s. 

Screen Shot 2020-05-20 at 11 20 47 AM

I've marked this PR as a WIP because: 

1. I'm unable to validate that this works with single-spa-playground. It seems to choke because the code there is attempting to parse the javascript as json. I feel like it would be best if it were able to validate there. 

Screen Shot 2020-05-20 at 11 26 24 AM

2. This should be deployed as an npm module. Ideally, this would be published as part of this repo's CI. For now, I've [hosted the files on my own now app](https://dist-qqbprggud.now.sh/single-spa-welcome.js)

3. There are still a few additional changes (eg. tests, maybe removing react and instead using single-spa-html, updating the import url once deployed, etc.) but I figured I'd get feedback on the direction things are going first.


packages/generator-single-spa/src/root-config/templates/src/activity-functions.js
```diff
@@ -1,4 +1,3 @@
-export function navbar(location) {
-  // The navbar is always active
+export function always(location) {
```
filoxo, line 1: This seemed more generic and easier to understand (for me).

packages/generator-single-spa/src/root-config/templates/src/index.ejs
```diff
@@ -57,6 +57,7 @@
   <% } %>
 </head>
 <body>
+  <main></main>
```
filoxo, line 60: This is needed because otherwise when mounting an app to `document.body` it removes the elements underneath it (system.import, etc.).

packages/generator-single-spa/src/root-config/templates/src/root-config.js
```diff
@@ -2,11 +2,18 @@ import { registerApplication, start } from "single-spa";
 import * as isActive from "./activity-functions";
 
 registerApplication({
-  name: "@<%- orgName %>/navbar",
-  app: () => System.import("@<%- orgName %>/navbar"),
-  activeWhen: isActive.navbar,
+  name: "@single-spa/welcome",
+  app: () =>
+    System.import("https://dist-qqbprggud.now.sh/single-spa-welcome.js"),
```
filoxo, line 7: TODO: Update with production url once deployed.

packages/single-spa-welcome/src/welcome.test.js
```diff
@@ -0,0 +1,10 @@
+import React from "react";
+import { render } from "@testing-library/react";
+import Welcome from "./welcome";
+
+describe("Welcome component", () => {
```
filoxo, line 5: TODO: Update or remove tests

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -48,8 +48,8 @@ function webpackConfigSingleSpa(opts) {
           },
         },
         {
-          test: /\.css$/,
-          include: /node_modules/,
+          test: /\.css$/i,
+          include: [/node_modules/, /src/],
```
filoxo, line 52: This is needed in order to be able to import local css files.

Sauloxd:
I'm not sure what's going on in the playground, but in the step you printed the error, it tries to do the following:
- Download your bundle not as JS but as text
- Run a regex to match the following: `System.register([MATCH THIS ARRAY], function() ...`
- JSON.parse the match, and assumes this array is the dependencies of your bundle
- Validates if your dependencies is already present in playground/import-map-overrides

Maybe the regex there is broken there, will check later 🤔

packages/generator-single-spa/src/root-config/templates/src/root-config.js
joeldenning: Could we just use unpkg instead of now.sh? That way a publish to npm would result in a deploy?

```suggestion
    System.import("https://unpkg.com/single-spa-welcome/dist/single-spa-welcome.js"),
```

packages/generator-single-spa/src/root-config/templates/src/activity-functions.js
joeldenning: Can't this entire file be deleted now that we're using the object version of registerApplication?

packages/generator-single-spa/src/root-config/templates/src/root-config.js
```diff
@@ -2,11 +2,18 @@ import { registerApplication, start } from "single-spa";
...
+  activeWhen: isActive.always,
 });
 
+// registerApplication({
+//   name: "@<%- orgName %>/navbar",
+//   app: () => System.import("@<%- orgName %>/navbar"),
+//   activeWhen: isActive.always,
```
joeldenning, line 14:
```suggestion
//   activeWhen: ['/']
```

packages/single-spa-welcome/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "husky": "^3.1.0",
+    "identity-obj-proxy": "^3.0.0",
+    "jest": "^24.9.0",
+    "jest-cli": "^24.9.0",
+    "prettier": "^2.0.4",
+    "pretty-quick": "^2.0.1",
+    "single-spa-react": "^2.14.0",
+    "style-loader": "^1.0.1",
+    "systemjs-webpack-interop": "^1.1.2",
+    "ts-config-single-spa": "^1.7.0",
```
joeldenning, line 44: Hmm was this generated from create-single-spa? Looks like we are unintentionally installing this dependency even on javascript projects? I'll create an issue for this.

joeldenning: https://github.com/single-spa/create-single-spa/issues/125

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
joeldenning: 👍

packages/single-spa-welcome/src/welcome.js
```diff
@@ -0,0 +1,105 @@
...
+          className="logo"
+          src="https://single-spa.js.org/img/logo-white-bgblue.svg"
+          alt=""
+        />
+        <div>
+          <h2 style={{ marginBottom: 0 }}>
+            Welcome{" "}
+            <span style={{ display: "block", fontSize: "1rem" }}>
+              to your single-spa root config!{" "}
+              <span role="img" aria-label="Congrats!">
```
joeldenning, line 18: #a11ypro

(nice work 😄)

packages/single-spa-welcome/src/welcome.js
```diff
@@ -0,0 +1,105 @@
...
+      <ul>
+        <li>
+          <a href="https://single-spa.js.org/docs/recommended-setup/#shared-dependencies">
+            Shared dependencies documentation on single-spa.js.org
+          </a>
+        </li>
+        <li>
+          <a href="https://github.com/systemjs/systemjs/">SystemJS</a> and{" "}
+          <a href="https://github.com/systemjs/systemjs/blob/master/docs/import-maps.md">
+            importmaps
```
joeldenning, line 76:
```suggestion
            Import Maps
```

packages/generator-single-spa/src/root-config/templates/src/activity-functions.js
filoxo: I like it!

packages/generator-single-spa/src/root-config/templates/src/root-config.js
filoxo: Yep! I'll just need to work with whatever the CI process is to deploy it as its own package. Is that just a lerna command?

packages/single-spa-welcome/package.json
filoxo: Yes, this was a fresh install. I noticed this but forgot to log an issue.

filoxo:
@Sauloxd yea, looks like the regex isn't matching correctly. Here's what the matches capture: 

Screen Shot 2020-05-20 at 1 55 26 PM

Perhaps its because its an empty array?

packages/generator-single-spa/src/root-config/templates/src/root-config.js
joeldenning: `yarn publish-all` does this. It calls `lerna publish`, which will publish all the packages in the monorepo at once

joeldenning:
single-spa-playground fix for this at https://github.com/single-spa/single-spa-playground/pull/25

packages/single-spa-welcome/src/welcome.test.js
filoxo: This test actually still works, and is probably better to keep than not.

filoxo:
I was able to test this successfully on single-spa-playground with the new fix added.

filoxo:
Oh I think was already updated in this commit https://github.com/single-spa/create-single-spa/pull/124/commits/7c83f034b9a95d749c5a7c9d06bed4ede3d5abb8

joeldenning:
Oops I missed that!

PR #126, opened 2020-05-20
Do not install TS dependencies in JS templates

### LINKED ISSUE #125: react javascript project installs ts-config-single-spa
See https://github.com/single-spa/create-single-spa/pull/124/files#r428225987

filoxo:
Resolves #125 

I noticed that these TS dependencies were also in the root-config and utility-module templates.


packages/generator-single-spa/src/react/templates/package.json
```diff
@@ -22,7 +22,6 @@
     "@babel/runtime": "^7.8.7",
     "@testing-library/jest-dom": "^5.5.0",
     "@testing-library/react": "^9.4.0",
-    "@types/jest": "^24.0.23",
```
joeldenning, line 25: I actually like having `@types/jest` and `@types/systemjs` installed for non typescript projects, because it provides autocomplete when i do `describe()` and `System.import()` inside of the IDE. VS Code will try to autocomplete those things incorrectly otherwise.

I understand that that's a weird reason to have a dependency, though, so I'm okay with removing this and suffering through worse autocomplete :)

filoxo: That makes total sense! Should that be for both `jest` and `systemjs` types?

PR #129, opened 2020-05-26
Setting babel_env to target current node during tests. Resolves #128.

### LINKED ISSUE #128: Creating a single spa app in typescript has some problems
Hi all,

I created both the root config and a react app with this tool in typescript and I found some problems.

- the tests don't work if you import something from another file 
	- this is because the correct transformer for jest should be `ts-jest`
- eslint display the error `  24:31  error  'X' is not defined no-undef` when using typescript specific type declaration (`type` and `interface`)
	- this one I didn't investigate too much I simply disabled lint in husky check

here is a repo created for the occasion https://github.com/zanza00/single-spa-testing-problem

I tried to fix the error by presenting a PR but I am not very familiar with the codebase and didn't found where to fix the issue, If someone is able to help me navigate the code I can do the PR myself.

joeldenning:
See #128

### DIFF
--- a/packages/generator-single-spa/src/react/templates/package.json
+++ b/packages/generator-single-spa/src/react/templates/package.json
@@ -6,8 +6,8 @@
     "lint": "eslint src --ext js<% if (typescript) { %>,ts,tsx<% } %>",
     "format": "prettier --write './**'",
     "test": "cross-env BABEL_ENV=test jest",
-    "watch-tests": "jest --watch",
-    "coverage": "jest --coverage"
+    "watch-tests": "cross-env BABEL_ENV=test jest --watch",
+    "coverage": "cross-env BABEL_ENV=test jest --coverage"
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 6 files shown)


packages/generator-single-spa/src/react/templates/package.json
```diff
@@ -5,7 +5,7 @@
     "analyze": "webpack --mode=production --env.analyze=true",
     "lint": "eslint src --ext js<% if (typescript) { %>,ts,tsx<% } %>",
     "format": "prettier --write './**'",
-    "test": "jest",
+    "test": "cross-env BABEL_ENV=test jest",
     "watch-tests": "jest --watch",
     "coverage": "jest --coverage"
```
zanza00, line 10: the BABEL_ENV is needed everywhere jest is used or am I missing something?

joeldenning: 👍 good catch, yes it is needed when running the coverage command, too. I will add it.

joeldenning: Updated

Issue #123, opened 2020-05-20
Create-single-spa of svelte from cli


limitall:
Please provide steps or add svelte as a option in cli

joeldenning:
Hi @limitall - sure let's keep this issue open to track it. The single spa core team might get to this at some point, and would also welcome outside contribution on it. All others interested in this, please show support with a thumbs up emoji on the original issue.

tangkunyin:
@limitall The cli did't provide svelte template until version 1.9.0.But you can refer this: https://single-spa.js.org/docs/ecosystem-svelte

Refer to my operation below：

1. create a svelte app
```
npx degit sveltejs/template svelte-app
```

2. create index.js like this
```
import singleSpaSvelte from "single-spa-svelte";
import SvelteApp from "./main";

const svelteLifecycles = singleSpaSvelte({
  component: function () {
    return SvelteApp;
  },
  domElementGetter: () => document.getElementById("svelte-app"),
  props: {},
});

export const bootstrap = svelteLifecycles.bootstrap;
export const mount = svelteLifecycles.mount;
export const unmount = svelteLifecycles.unmount;
```

3. registe application in root-config
```
registerApplication({
  name: "svelte-app",
  app: () => System.import("@your-org/svelte-app"),
  activeWhen: isActive.svelte,
});
```

4. import map

```
<script type="systemjs-importmap">
    {
      "imports": {
        "@your-og/svelte-app": "//localhost:9003/build/bundle.js"
      }
    }
  </script>
```

filoxo:
Generating a Svelte application was added to [create-single-spa@1.10.0](https://github.com/single-spa/create-single-spa/releases/tag/v1.10.0).

PR #132, opened 2020-05-27
Feat/add svelte template

filoxo:
Beginning work for resolving #123, which is generating a Svelte app. At the moment it doesn't include TypeScript support or testing.


packages/generator-single-spa/src/svelte/generator-single-spa-svelte.js
```diff
@@ -0,0 +1,178 @@
...
+
+    const srcFileExtension = /* this.options.typescript ? "tsx" : */ "js";
+
+    this.fs.copyTpl(
+      this.templatePath("../../common-templates/gitignore"), // this is relative to /templates
+      this.destinationPath(".gitignore"),
+      this.options
+    );
+    // this.fs.copyTpl(
+    //   this.templatePath(".prettierignore"),
```
joeldenning, line 112: Are you planning on adding prettier to the svelte generator? Does prettier know how to handle svelte files?

packages/generator-single-spa/src/svelte/templates/rollup.config.js
```diff
@@ -0,0 +1,70 @@
+import svelte from "rollup-plugin-svelte";
```
joeldenning, line 1: Did you write this rollup config? What did you base it on?

packages/generator-single-spa/src/svelte/templates/yarn.lock
```diff
@@ -0,0 +1,583 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
joeldenning, line 1: yarn.lock files should not be templated - this one can be removed

packages/generator-single-spa/src/svelte/generator-single-spa-svelte.js
filoxo: Prettier is supported though a plugin https://github.com/sveltejs/prettier-plugin-svelte

packages/generator-single-spa/src/svelte/templates/rollup.config.js
filoxo: This config is based on https://github.com/sveltejs/template

packages/generator-single-spa/src/svelte/templates/yarn.lock
filoxo: Ah, good call

packages/generator-single-spa/src/svelte/templates/rollup.config.js
joeldenning: 👍

joeldenning:
Another thing - could we add an e2e test for this? Also have you verified by running locally that this produces a project that works?

filoxo:
Yea, I need to add e2e tests as well. I realized that as I opened up the PR 🙉 But yes, I did run it locally and it renders an app that I validated with a local root config as well as single-spa-playground.

PR #134, opened 2020-05-29
Resolve Webpack config loaders as dependencies

### LINKED ISSUE #83: Move babel-loader, css-loader, and style-loader into webpack-config-single-spa
Instead of peerDependencies, these could be dependencies. To do so would require code like this:

```js
{
  modules: {
    rules: [
      { loader: require.resolve('babel-loader', {paths: [__dirname]}) }
    ]
  }
}
```

filoxo:
Resolves #83


packages/webpack-config-single-spa/package.json
```diff
@@ -24,9 +24,7 @@
   "dependencies": {
     "clean-webpack-plugin": "^3.0.0",
     "unused-files-webpack-plugin": "^3.4.0",
-    "webpack-bundle-analyzer": "^3.6.0"
-  },
-  "peerDependencies": {
+    "webpack-bundle-analyzer": "^3.6.0",
     "babel-loader": "*",
```
joeldenning, line 28: Let's specify exact version now, instead of `*`? The star was to try to avoid situations where the user-provided version was incorrect and that resulting in a broken build. However, now that the user doesn't have to provide the version anymore, I don't think that's a problem?

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -44,16 +44,18 @@ function webpackConfigSingleSpa(opts) {
           test: /\.(js|ts)x?$/,
           exclude: /node_modules/,
           use: {
-            loader: "babel-loader",
+            loader: require.resolve("babel-loader", { paths: [__dirname] }),
```
joeldenning, line 47: We should remove these as devDependencies from the generator package.json templates now, since it's baked into webpack-config-single-spa

packages/webpack-config-single-spa/package.json
filoxo: Good call. I'll add the exact versions.

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
filoxo: I should have thought about that before. I agree.

Issue #94, opened 2020-04-17
Add console instructions for Angular/Vue to test in single-spa-playground


joeldenning:
Similar to the languaging in the react generator, I think we could encourage users to try the "instant test" feature in the single-spa playground

filoxo:
This is resolved and published in [v1.11.0](https://github.com/single-spa/create-single-spa/releases/tag/v1.11.0).

PR #137, opened 2020-06-05
Feat/additional console instruction vue

filoxo:
Partial work for #94 by adding instant-test instructions for newly generated Vue applications.

### DIFF
--- a/packages/generator-single-spa/src/vue/generator-single-spa-vue.js
+++ b/packages/generator-single-spa/src/vue/generator-single-spa-vue.js
@@ -4,6 +4,27 @@ const util = require("util");
 const commandExists = util.promisify(require("command-exists"));
 
 module.exports = class SingleSpaVueGenerator extends Generator {
+  constructor(args, opts) {
+    super(args, opts);
+
+    this.option("packageManager", {
+      type: String,
+    });
+  }
+  async getOptions() {
+    if (!this.options.packageManager) {
+      this.options.packageManager = (
+        await this.prompt([
+          {
+            type: "list",
+            name: "packageManager",
+            message: "Which package manager do you want to use?",
+            choices: ["yarn", "npm"],
+          },
+        ])
+      ).packageManager;
+    }
+  }
   async runVueCli() {
     const globalInstallation = await commandExists("vue");
 
--- a/packages/generator-single-spa/src/vue/generator-single-spa-vue.js
+++ b/packages/generator-single-spa/src/vue/generator-single-spa-vue.js
@@ -3,6 +3,7 @@ const { spawnSync } = require("child_process");
 const util = require("util");
 const commandExists = util.promisify(require("command-exists"));
 const chalk = require("chalk");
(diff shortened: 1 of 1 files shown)


filoxo:
Related: https://github.com/single-spa/vue-cli-plugin-single-spa/pull/10 fixes an issue where assets and chunks won't load because of the structure of default Vue applications.

packages/generator-single-spa/src/vue/generator-single-spa-vue.js
```diff
@@ -2,8 +2,30 @@ const Generator = require("yeoman-generator");
 const { spawnSync } = require("child_process");
 const util = require("util");
 const commandExists = util.promisify(require("command-exists"));
+const chalk = require("chalk");
 
 module.exports = class SingleSpaVueGenerator extends Generator {
+  constructor(args, opts) {
+    super(args, opts);
+
+    this.option("packageManager", {
```
joeldenning, line 11: If we do not provide this to Vue CLI as an argument, it will prompt the user for this. Why do we need to do it here?

Is it so that we can provide the correct final instructions for the instant test (`yarn serve` versus `npm run server`)?

filoxo: I'm not getting prompted for which package manager to use with \@vue/cli 4.4.1, but I'm realizing that its because at some point it saved it into my `~/.vuerc`. And yes, it is helpful to have that option before trigging the vue cli so we can show the corresponding command once it completes.

joeldenning: I generally would lean towards letting Vue CLI handle as many options as possible. That way there is less for create-single-spa to do, and also the UI we present to the developer is familiar to them if they have used Vue CLI.

Perhaps an alternative would be to check for the presence of a `yarn.lock` or `package-lock.json` file after Vue has finished? And then change the colored instructions based on that?

What do you think?

filoxo: Yea, that would work well! I'll change it.

packages/generator-single-spa/src/vue/generator-single-spa-vue.js
```diff
@@ -35,13 +37,28 @@ module.exports = class SingleSpaVueGenerator extends Generator {
...
-          VUE_CLI_SKIP_DIRTY_GIT_PROMPT: true
-        })
+          VUE_CLI_SKIP_DIRTY_GIT_PROMPT: true,
+        }),
       });
     }
   }
+
+  async finished() {
+    const usedYarn = this.fs.exists(path.resolve(this.cwd, "yarn.lock"));
```
filoxo, line 52: I chose to detect if yarn was used because it seems like it'd a more reliable check. I'm not familiar with other package managers (pnpm, cnpm) but I figure the base `npm` command would be generic enough that users could infer how to update the command themselves.

PR #143, opened 2020-06-30
Feat: generate a root-config with single-spa-layout

### LINKED ISSUE #141: Generate a root-config that includes single-spa-layout
- single-spa-layout is released as a beta
- Documentation is complete for single-spa-layout
- I think it would be good to include or at least add some commented code to show:
    - Loading UIs when applications are downloaded.
    - Default routes for Not Found / 404 pages.

For an example, see https://github.com/react-microfrontends/root-config/pull/11

This option should be clearly marked as "beta" when the user is prompted, as well as having multiple places with a link to open an issue/send feedback.

filoxo:
Resolves #141 

I tested this by generating a root-config with both JS and TS. They both run without failures, but I still need to write e2e tests.


packages/generator-single-spa/src/root-config/templates/src/layout/index.ejs
```diff
@@ -0,0 +1,90 @@
...
+  <script src="https://cdn.jsdelivr.net/npm/systemjs@6.3.1/dist/extras/amd.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/systemjs@6.3.1/dist/extras/named-exports.js"></script>
+  <% } else { %>
+  <script src="https://cdn.jsdelivr.net/npm/import-map-overrides@1.14.6/dist/import-map-overrides.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/systemjs@6.3.1/dist/system.min.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/systemjs@6.3.1/dist/extras/amd.min.js"></script>
+  <script src="https://cdn.jsdelivr.net/npm/systemjs@6.3.1/dist/extras/named-exports.min.js"></script>
+  <% } %>
+
+  <template id="single-spa-layout">
```
joeldenning, line 71: Instead of creating a new template file for this, could we put this `<template>` inside of an ejs `if` statement? That way we don't have to upgrade dependencies and duplicate changes in two places

packages/generator-single-spa/src/root-config/templates/webpack.config.js
```diff
@@ -20,6 +21,7 @@ module.exports = (webpackConfigEnv) => {
         template: "src/index.ejs",
         templateParameters: {
           isLocal: webpackConfigEnv && webpackConfigEnv.isLocal === "true",
+          orgName,
```
joeldenning, line 24: Interesting - I like it! Automatically adding the org name there.

packages/generator-single-spa/src/root-config/templates/src/layout/index.ejs
filoxo: Yea this was tricky... because its not possible to partially evaluate an ejs file. Let me see if I can explain what I mean... create-single-spa in some instances uses ejs files to apply some changes to the final output. But in this case, it would need to use an ejs file to generate an ejs file (since the final output needs to be an ejs file to work with HtmlTemplatePlugin) which might mean double escaping or some other weirdness.

filoxo: I guess we could instead put both layouts in the same template and control it with some boolean in the webpack config. I just think that would be confusing and leads to bloat (it would still install single-spa-layout even though someone may not want to use it).

joeldenning: Ah good points I didn't think that through. The ejs template is processed after generation, not before. Makes sense.

joeldenning:
I am not sure why the tests failed in the build. Will try to restart them.

filoxo:
Not sure what's up with these tests. They all pass on my machine and don't take 20 mins 😭 

Screen Shot 2020-07-02 at 11 54 13 AM

joeldenning:
I'm okay with merging this as-is and chalking it up to flaky tests. Feel free to do so and publish a new version.

filoxo:
I have one more hunch I'll try chasing down.

joeldenning:
Hmm perhaps we should just merge this? And deal with the tests in a separate PR?

PR #145, opened 2020-07-07
Fix/missing org name

filoxo:
If no orgName or projectName is provided or spaces are included, then the output has some weird references and filenames. These changes guard against providing empty and invalid input, validating that the names match the "use lowercase and dashes" instructions.

These changes also reorganize the locations of where options are being prompted for so that they are in the same function block.

### DIFF
--- a/packages/generator-single-spa/package.json
+++ b/packages/generator-single-spa/package.json
@@ -8,7 +8,7 @@
     "src"
   ],
   "scripts": {
-    "test": "jest --testPathIgnorePatterns templates"
+    "test": "jest"
   },
   "version": "1.11.0",
   "main": "src/generator-single-spa.js",
@@ -22,5 +22,10 @@
     "jest": "^24.9.0",
     "yeoman-assert": "^3.1.1",
     "yeoman-test": "^2.0.0"
+  },
+  "jest": {
+    "testPathIgnorePatterns": [
+      "templates"
+    ]
   }
 }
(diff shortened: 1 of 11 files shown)


packages/generator-single-spa/package.json
```diff
@@ -8,7 +8,7 @@
     "src"
   ],
   "scripts": {
-    "test": "jest --testPathIgnorePatterns templates"
```
filoxo, line 11: I moved this to the jest config to support being able to run `yarn test basic.test.js`. Without it, it would actually append that arg to the `testPathIgnorePatterns` and it would actually run all the other tests instead of just the one you wanted.

packages/generator-single-spa/src/react/generator-single-spa-react.js
```diff
@@ -20,7 +21,7 @@ module.exports = class SingleSpaReactGenerator extends Generator {
       type: String,
     });
   }
-  async createPackageJson() {
+  async getOptions() {
```
filoxo, line 24: It helped to organize these into a separate getOptions block.

packages/generator-single-spa/src/react/generator-single-spa-react.js
```diff
@@ -47,6 +48,41 @@ module.exports = class SingleSpaReactGenerator extends Generator {
...
+
+      orgName = orgName && orgName.trim();
+      if (!orgName) console.log(chalk.red("orgName must be provided!"));
+      if (!isValidName(orgName))
+        console.log(chalk.red("orgName must use lowercase and dashes!"));
+      this.options.orgName = orgName;
+    }
+
+    while (!this.options.projectName) {
+      let { projectName } = await await this.prompt([
```
filoxo, line 68: TYPO: await await

PR #146, opened 2020-07-07
Check and notify user if a newer version is available

### LINKED ISSUE #142: Add a notification banner for outdated version of create-single-spa
Feature request: 
I was helping someone recently with a create-single-spa question and I made the mistake of assuming they had the latest version (see https://single-spa.slack.com/archives/C8R6U7MT7/p1592326497247200?thread_ts=1592320975.228600&cid=C8R6U7MT7). If someone instead installed via `npm i -g create-single-spa` then it would get out of date after some time.
I think it would be great to show a notification if the user is not using the latest version of create-single-spa and prompt them to update. Other projects such as yarn display something like this which I think is helpful. Here is an example from another package I have globally installed:

Screen Shot 2020-06-19 at 10 56 33 AM

I can get to this eventually but this would be a great first feature if someone would like to contribute. Feel free to comment if you're interested in taking this on, and we would be happy to help/answer any questions.

filoxo:
Resolves #142 

Below is a screenshot of what the terminal message with colors when an outdated version is detected.

Screen Shot 2020-07-07 at 4 37 16 PM

I decided last minute to add the underline styling (to help distinguish from the rest of the prompt) after I took the above screenshot.

Screen Shot 2020-07-07 at 4 45 44 PM

Feedback: 
- Should there be some sort of flag or config we can add to silence this message? 
- ~~I can't find a good way to test this... should I maybe move it into an external module to simplify testing?~~ I did move these into an external module with some corresponding tests.

### DIFF
--- a/packages/generator-single-spa/package.json
+++ b/packages/generator-single-spa/package.json
@@ -8,7 +8,7 @@
     "src"
   ],
   "scripts": {
-    "test": "jest --testPathIgnorePatterns templates"
+    "test": "jest"
   },
   "version": "1.11.0",
   "main": "src/generator-single-spa.js",
@@ -22,5 +22,10 @@
     "jest": "^24.9.0",
     "yeoman-assert": "^3.1.1",
     "yeoman-test": "^2.0.0"
+  },
+  "jest": {
+    "testPathIgnorePatterns": [
+      "templates"
+    ]
   }
 }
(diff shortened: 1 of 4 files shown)


filoxo:
Here is an updated screenshot of the update messages. These lines are generated in tests, but will be the same if run with the cli; it was just easier to output in the test environment.

Screen Shot 2020-07-09 at 9 52 11 AM

joeldenning:
Excited to see this

PR #150, opened 2020-07-14
Fixing "format" script on Windows.

joeldenning:
On Windows, the `npm run format` command doesn't work.

Before fix:

image

After fix:

image


joeldenning:
I have confirmed that the fix also still works on Mac

filoxo:
🙈

PR #156, opened 2020-07-20
Fix/vue app orgname

### LINKED ISSUE #148: Vue apps do not prompt for organization name
When you use create-single-spa to generate a vue app, you are not prompted for the organization name, which is added as a prefix `@org/`. See https://github.com/single-spa/vue-cli-plugin-single-spa/blob/8b7076be0b8cb45c6596a5a6ac283c65e6929834/template/src/set-public-path.js#L3 and https://github.com/single-spa/vue-cli-plugin-single-spa/blob/58caee10211696d2c7a8780d6eb6c21a46c2fd49/generator.js#L3.

The `appName` variable is coming from the package.json of the generated vue cli project. The package.json name comes from the argument passed to `vue create <dir name>`. More documentation about this at https://cli.vuejs.org/guide/creating-a-project.html#vue-create.

cc @mattjonesorg and @tylerkeesling who were there when we discovered this.

Here are proposed changes:

1. Add the organization and project name prompts to create-single-spa, similar to how most of the generators do it. [Example](https://github.com/single-spa/create-single-spa/blob/74a99b37c0ea3229216a82332faa0f34086c0bfb/packages/generator-single-spa/src/util-module/generator-single-spa-util-module.js#L54-L84)
2. Research if it's possible to pass in a name with `/` in it to Vue CLI without it creating a directory for that slash. If so, invoke Vue CLI in that way. [Code to change](https://github.com/single-spa/create-single-spa/blob/74a99b37c0ea3229216a82332faa0f34086c0bfb/packages/generator-single-spa/src/vue/generator-single-spa-vue.js#L29)

filoxo:
Resolves #148

The approach I took makes the following changes: 
- I made the `dir` a little more flexible and resilient by allowing it to be a folder path (eg. `create-single-spa ../my-new-app`)
- The generator now prompts for and validates the `orgName`
- `projectName` is derived from the `dir` (eg. from the above command the projectName would be `my-new-app` which is used when calling `vue create`)
    - `projectName` is now also validated like all the other options
- After the project is created, the package.json is modified to include the orgName (eg. `@org/my-new-app`)
- Then `vue add single-spa` is called, and the correct value is used by the plugin to include the orgName throughout

There are no tests for the Vue functionality because we need to invoke the \@vue/cli, and I'm not sure how to approach validating this beyond manual testing. Also, I'd like for someone to run this on a Windows machine in case there are any problems with the path.resolve calls.

### DIFF
--- a/packages/generator-single-spa/src/vue/generator-single-spa-vue.js
+++ b/packages/generator-single-spa/src/vue/generator-single-spa-vue.js
@@ -4,6 +4,7 @@ const util = require("util");
 const commandExists = util.promisify(require("command-exists"));
 const chalk = require("chalk");
 const path = require("path");
+const isValidName = require("../naming");
 
 module.exports = class SingleSpaVueGenerator extends Generator {
   constructor(args, opts) {
@@ -46,12 +47,17 @@ module.exports = class SingleSpaVueGenerator extends Generator {
       command += ".cmd";
     }
 
-    this.cwd = this.options.dir || ".";
+    // Derrive projectName value
+    const { dir, name } = path.parse(this.options.dir || ".");
+    if (!name) throw new Error("projectName must be provided!");
+    if (!isValidName(name))
+      throw new Error("projectName must use lowercase and dashes!");
 
     const { status, signal } = spawnSync(
       command,
-      args.concat(["create", this.cwd, "--skipGetStarted"]),
+      args.concat(["create", name, "--skipGetStarted"]),
       {
+        cwd: dir,
         stdio: "inherit",
       }
     );
--- a/packages/generator-single-spa/src/vue/generator-single-spa-vue.js
+++ b/packages/generator-single-spa/src/vue/generator-single-spa-vue.js
@@ -4,6 +4,7 @@ const util = require("util");
 const commandExists = util.promisify(require("command-exists"));
 const chalk = require("chalk");
 const path = require("path");
+const fs = require("fs");
 const isValidName = require("../naming");
 
 module.exports = class SingleSpaVueGenerator extends Generator {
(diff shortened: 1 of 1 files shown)


filoxo:
I'll post testing instructions/scenarios later today.

packages/generator-single-spa/src/vue/generator-single-spa-vue.js
```diff
@@ -22,12 +48,17 @@ module.exports = class SingleSpaVueGenerator extends Generator {
       command += ".cmd";
     }
 
-    this.cwd = this.options.dir || ".";
+    // Derrive projectName value
```
joeldenning, line 51:
```suggestion
    // Derive projectName value
```

packages/generator-single-spa/src/vue/generator-single-spa-vue.js
```diff
@@ -4,8 +4,34 @@ const util = require("util");
...
+          type: "input",
+          name: "orgName",
+          message: "Organization name (use lowercase and dashes)",
+        },
+      ]);
+
+      orgName = orgName && orgName.trim();
+      if (!orgName) console.log(chalk.red("orgName must be provided!"));
+      if (!isValidName(orgName))
+        console.log(chalk.red("orgName must use lowercase and dashes!"));
```
joeldenning, line 31: Is this meant to be a console warning, or an error thrown? The equivalent if statement for projects throws an error.

filoxo: It's meant to be a console warning. It will display that line and prompt the user again until it passes validation. If it were to throw an error instead, the process would exit which I would think is undesirable since it can be handled gracefully.

Can you share an example of the equivalent if statement are you thinking of? I had changed the orgName and projectName prompts throughout to work as I described above in PR #145.

filoxo:
> Regarding Vue CLI in the tests, wouldn't things still work since it runs npx when there is no global installation of Vue CLI?

That's what I thought but the tests seem to hang instead. I had tried to set up an e2e test but it would always hang for some reason. I'll investigate a little further.

packages/generator-single-spa/src/vue/generator-single-spa-vue.js
joeldenning: https://github.com/single-spa/create-single-spa/pull/156/files#diff-4bf5d50eb100da1f8c558788d5ac79bdR55

filoxo: I see. That is a slightly different case because dir is an "option" not a "prompt". The Vue app create is sightly different than for the others because the `projectName` is derived from the `dir` option; the others prompt for the `projectName`. Since its not a prompt, an error is the only way to handle this for Vue apps. I'm open to changing it so that this generator prompts for the projectName instead.

joeldenning: I see, makes sense 👍

joeldenning:
I'm okay with omitting the tests if they're being problematic.

filoxo:
OK finally circling back to this. Testing this programmatically has the following problems:
- unit testing within generator-single-spa isn't possible without mocking out the entire spawnSync call, and/or a filesystem to use, which I guess could be done with memfs but that's some effort and liable to break if/when the @vue/cli changes...
- and an e2e test isn't possible because after invoking the @vue/cli there are additional prompts that require interaction. In this case, when invoking the @vue/cli it expects you to select from options and there isn't a good way to send keys through nixt.

Not sure there's a good solution for this but I guess we can tackle that one it becomes a problem.

PR #165, opened 2020-08-10
Upgrading single-spa and systemjs-webpack-interop.

joeldenning:
Both libraries have IE11 fixes, and always good to keep them all up-to-date. See https://github.com/joeldenning/systemjs-webpack-interop/releases/tag/v2.1.2

### DIFF
--- a/packages/generator-single-spa/src/react/templates/package.json
+++ b/packages/generator-single-spa/src/react/templates/package.json
@@ -39,7 +39,7 @@
     "prettier": "^2.0.4",
     "pretty-quick": "^2.0.1",
     "single-spa-react": "^2.14.0",
-    "systemjs-webpack-interop": "^1.1.2",
+    "systemjs-webpack-interop": "^2.1.2",
     "webpack": "^4.41.2",
     "webpack-cli": "^3.3.10",
     "webpack-config-single-spa-react": "^1.0.3",
--- a/packages/generator-single-spa/src/root-config/templates/src/index.ejs
+++ b/packages/generator-single-spa/src/root-config/templates/src/index.ejs
@@ -32,10 +32,11 @@
   <script type="systemjs-importmap">
     {
       "imports": {
-        "single-spa": "https://cdn.jsdelivr.net/npm/single-spa@5.5.0/lib/system/single-spa.min.js"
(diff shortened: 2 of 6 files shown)


packages/generator-single-spa/src/root-config/templates/src/index.ejs
```diff
@@ -32,10 +32,11 @@
   <script type="systemjs-importmap">
     {
       "imports": {
-        "single-spa": "https://cdn.jsdelivr.net/npm/single-spa@5.5.0/lib/system/single-spa.min.js"
+        "single-spa": "https://cdn.jsdelivr.net/npm/single-spa@5.5.5/lib/system/single-spa.min.js"
       }
     }
   </script>
+  <link rel="preload" href="https://cdn.jsdelivr.net/npm/single-spa@5.5.5/lib/system/single-spa.min.js" as="script">
```
joeldenning, line 39: Minor perf boost

packages/generator-single-spa/src/react/templates/package.json
```diff
@@ -39,7 +39,7 @@
     "prettier": "^2.0.4",
     "pretty-quick": "^2.0.1",
     "single-spa-react": "^2.14.0",
-    "systemjs-webpack-interop": "^1.1.2",
+    "systemjs-webpack-interop": "^2.1.2",
```
joeldenning, line 42: See https://github.com/joeldenning/systemjs-webpack-interop/releases/tag/v2.0.0

packages/generator-single-spa/src/root-config/templates/src/index.ejs
filoxo: Should this also be added to the layout version of the template? https://github.com/single-spa/create-single-spa/blob/master/packages/generator-single-spa/src/root-config/templates/src/layout/index.ejs

joeldenning: It is - https://github.com/single-spa/create-single-spa/pull/165/files#diff-473cf7d0885b1dfb721b1a1067ed11e3R37

filoxo: Ah missed that. LGTM!

PR #172, opened 2020-08-28
Do not install react types in non-react projects. Resolves #171.

### LINKED ISSUE #171: Why React is Listed as Dependency by Default in new Single SPA app Generated w/CLI
**Describe the bug or question**
A clear and concise description of what the bug or question is.

I see this listed as dependencies in `src` `package.json`:

```
"dependencies": {
    "single-spa": "^5.3.1",
    "@types/jest": "^25.2.1",
    "@types/react": "^16.9.32",
    "@types/react-dom": "^16.9.6",
    "@types/systemjs": "^6.1.0",
    "@types/webpack-env": "^1.15.1"
  }
```

Why is React and React DOM listed as dependencies by default when creating a new app with Single SPA CLI ?

joeldenning:
See #171

### DIFF
--- a/packages/generator-single-spa/src/react/generator-single-spa-react.js
+++ b/packages/generator-single-spa/src/react/generator-single-spa-react.js
@@ -119,6 +119,15 @@ module.exports = class SingleSpaReactGenerator extends Generator {
         )
       );
 
+      this.fs.extendJSON(
+        this.destinationPath("package.json"),
+        this.fs.readJSON(
+          this.templatePath(
+            "../../common-templates/typescript/react.package.json"
+          )
+        )
+      );
+
       // Extend with react-specific package json for typescript
       this.fs.extendJSON(
         this.destinationPath("package.json"),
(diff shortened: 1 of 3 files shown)


joeldenning:
The build failures seem related to the svelte test, whose code is unchanged. I think the tests are brittle

packages/generator-single-spa/src/react/generator-single-spa-react.js
```diff
@@ -112,11 +112,20 @@ module.exports = class SingleSpaReactGenerator extends Generator {
     this.fs.extendJSON(this.destinationPath("package.json"), packageJson);
 
     if (this.options.typescript) {
+      const typescriptPackageJsonTemplate = await fs.readFile(
+        this.templatePath("../../common-templates/typescript/package.json.ejs"),
+        "utf-8"
+      );
+      const typescriptPackageJsonStr = ejs.render(
```
filoxo, line 119: Would it be simpler to have a separate `reactTypes.json` file instead? My thought process is that this isn't a true package.json (at least not one that contains everything necessary for an npm project), and it would avoid having to ejs render them. I'm failing to think of any benefits that I could be overlooking though.

joeldenning: sure

joeldenning: I have updated this with your feedback

PR #181, opened 2020-09-11
Add license file. Resolves #180.

### LINKED ISSUE #180: Missing license
When publishing the latest version of create-single-spa, I noticed the following message: 

```sh
lerna WARN ENOLICENSE Packages create-single-spa and generator-single-spa are missing a license.
lerna WARN ENOLICENSE One way to fix this is to add a LICENSE.md file to the root of this repository.
lerna WARN ENOLICENSE See https://choosealicense.com for additional guidance.
lerna success published generator-single-spa 1.12.6
```

And indeed the top level package as well as the subpackages do not have a license file. We should probably add the same one used with other single-spa helpers.

joeldenning:
See #180. Lerna automatically will copy this license file into all other packages during publish. More info at https://github.com/lerna/lerna/issues/1666

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+License (MIT)
+
+Copyright (c) 2020 single-spa
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


filoxo (approved):
🙈

PR #182, opened 2020-09-17
Feat/allow numbers in naming

### LINKED ISSUE #177: Cannot use numbers either in org or project name
Hello,

I met an issue following a tutorial here: https://dev.to/thawkin3/developing-and-deploying-micro-frontends-with-single-spa-39ij

I think it's normal to use numbers in project name.

For example, org name like `back2warcraft`,  project name like `single-spa-demo-1` should be OK. 

I found probably it's becoz that cli `v1.12.3` made some breaking change? We cannot include numbers in project name any more.

See this line: https://github.com/single-spa/create-single-spa/blob/59d29ec41e2848a0e6c1b932bcc66504079c4127/packages/generator-single-spa/src/naming.js#L1

filoxo:
Resolves #177 

I learned while doing this that the prompt lib that Yeoman uses actually supports validation directly, so with this I was able to change how the input is validated rather than using a while loop. I'm going to have a follow-up PR to further clean up the prompts code.


packages/generator-single-spa/src/validate-naming.js
```diff
@@ -0,0 +1,15 @@
+const BEGIN_WITH_LETTER = /^[a-z]/;
+const ALLOWED_CHARACTERS = /^[a-z0-9\-]+$/;
```
joeldenning, line 2: should we allow capital characters, too?

filoxo: I debated that as well. I don't see anything wrong with it since 1) for single-spa its mostly a convention rather than a requirement, and 2) that is permitted in package.json naming. The only counterargument is that its gone this long without being something that users have requested/reported. But it should be pretty easy to add if we do want it.

joeldenning: Yeah I say let's add it - no reason to enforce lowercase i think

filoxo:
I had mentioned that I had "a follow-up PR to further clean up the prompts code". I went ahead and pulled it into this PR to fix an obnoxious problem that I introduced. Writing it down here: 

```
this.options.orgName = (
      await this.prompt([
        {
          type: "input",
          name: "orgName",
          message: "Organization name",
          suffix: " (can use lowercase letters, numbers, dash or underscore)",
          validate,
        },
      ]);
  ).orgName;
```

This is bad because the prompt was never running, and then overriding the `option.orgName` with `undefined`. With batching the prompts, I can merge them all into `this.options` and if a prompt was skipped it will not override the existing value.

packages/generator-single-spa/src/vue/generator-single-spa-vue.js
```diff
@@ -20,49 +20,34 @@ module.exports = class SingleSpaVueGenerator extends Generator {
...
-          console.log(chalk.red("projectName must use lowercase and dashes!"));
-        } else {
-          this.options.projectName = projectName;
-        }
-      }
-    } else {
-      this.options.dir = dir;
-      this.options.projectName = name;
+    if (!fs.existsSync(this.options.dir)) {
+      fs.mkdirSync(this.options.dir);
```
filoxo, line 50: This fixed a bug I encountered where the vue-cli was failing silently if the directory did not exist.

joeldenning (approved):
lgtm

PR #184, opened 2020-09-20
Adding single-spa-web-server-utils package

joeldenning:
I decided not to put this logic directly into import-map-overrides or single-spa-layout, since it's sort of outside of the scope of those packages. I envision the single-spa-web-server-utils package growing to include any opinionated / recommended logic to add to web servers. Including possibly an entire request handler so that generated root configs that want to do server rendering defer all / most of their logic to this package, which then calls into single-spa-layout and import-map-overrides to do stuff.


filoxo (approved):
I agree that it probably doesn't fit in those repos. I almost feel like this repo should be migrated into a general "single-spa utils" repo rather than create-single-spa as the top-level name. But that doesn't have to happen anytime soon yet. 

🙈

joeldenning:
Yeah I wondered about whether it fits in create-single-spa monorepo or not. I think that it might, since it's not required for using ssr + single-spa, but is part of the recommendation. I guess we'll see over time if that ends up being a good choice.

PR #188, opened 2020-09-29
Remove packageManager prompt for Angular apps

### LINKED ISSUE #186: Error when @angular/cli@8 is being used
The issue was raised on Slack. See https://single-spa.slack.com/archives/C8R6U7MT7/p1601182403067100

When @angular/cli@8 is installed locally (and presumably versions below), create-single-spa errors with:

```sh
Unknown option: --package-manager
Unknown option: yarn.
```

Screenshot 2020-09-27 at 10 16 35 AM

If you run `npx @angular/cli@8.3.29 new --help` you'll see that it doesn't have an option for `--package-manager` but the next version does (as verified by running `npx @angular/cli@9 new --help`).

filoxo:
Resolves #186 by simply removing the prompt and not passing through the packageManager option to the @angular/cli. I preferred this because 1) it is quite a bit of work to extract out the angular cli version, and 2) should be better for future versions since they may remove or change it (and then c-s-s would be forced to do yet another check).

### DIFF
--- a/packages/generator-single-spa/src/angular/generator-single-spa-angular.js
+++ b/packages/generator-single-spa/src/angular/generator-single-spa-angular.js
@@ -9,23 +9,12 @@ module.exports = class SingleSpaAngularGenerator extends Generator {
   constructor(args, opts) {
     super(args, opts);
 
-    this.option("packageManager", {
-      type: String,
-    });
-
     this.option("projectName", {
       type: String,
     });
   }
   async getOptions() {
     const answers = await this.prompt([
-      {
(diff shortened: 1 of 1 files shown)


joeldenning:
Does angular prompt for package manager in CLI 9/10? Or does omitting it make it default to npm?

filoxo:
Correct, omitting it defaults to npm. Its not a required flag for the angular/cli.

joeldenning (approved):
Hmm I wish we could force Angular to prompt for it. But this is good for now.

PR #190, opened 2020-10-02
single-spa-layout is no longer in beta

joeldenning:
See https://github.com/single-spa/single-spa-layout/releases/tag/v1.0.1


filoxo (approved):
🙈 👏

PR #191, opened 2020-10-02
Feature: dynamically show instructions for single-spa-layout root-configs

filoxo:
In order to better support single-spa-layout and improve a beginner's experience, I'm adding a feature that detects if a page has single-spa-router and displaying different "next steps" based on that. 

I also made some minor changes to the content to explain that shared dependencies are required for a React application.

### DIFF
--- a/packages/single-spa-welcome/src/welcome.js
+++ b/packages/single-spa-welcome/src/welcome.js
@@ -7,7 +7,7 @@ export default function Root(props) {
     // While this won't work for SSR, the goal of this app is to help
     // beginners, which likely won't use SSR before running their
     // first root-config.
-    const templates = Array.prototype.slice.call(
+    const templates = Array.from(
       document.querySelectorAll("template")
     );
 
--- a/packages/single-spa-welcome/src/welcome.js
+++ b/packages/single-spa-welcome/src/welcome.js
@@ -1,7 +1,24 @@
-import React from "react";
+import React, { useState } from "react";
 import "./welcome.css";
 
 export default function Root(props) {
+  const [isUsingSingleSpaLayout] = useState(() => {
+    // Search the dom for a single-spa-router element
+    // While this won't work for SSR, the goal of this app is to help
+    // beginners, which likely won't use SSR before running their
+    // first root-config.
+    const templates = Array.prototype.slice.call(
+      document.querySelectorAll("template")
+    );
+
+    return templates.some((template) => {
+      const hasSingleSpaRouter = !!template.content.querySelector(
+        "single-spa-router"
+      );
+      return hasSingleSpaRouter;
+    });
+  });
+
   return (
     <section id="welcome">
       <div className="banner">
@@ -30,7 +47,7 @@ export default function Root(props) {
       <h3>1. Add shared dependencies</h3>
       <ul>
         <li>
-          Locate the importmap in <code>src/index.ejs</code>
+          Locate the import map in <code>src/index.ejs</code>
         </li>
         <li>
           <p>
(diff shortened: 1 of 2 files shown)


filoxo:
I tested this by loading single-spa-welcome to single-spa-playground (which uses the `registerApplication` api) and react.microfrontends.app (which uses single-spa-layout). Here is a screenshot of each 

Screen Shot 2020-10-02 at 2 20 57 PM

Note: the styling differences below are due to the global css applied to react.microfrontends.app.

packages/single-spa-welcome/src/welcome.js
```diff
@@ -1,7 +1,24 @@
...
+      document.querySelectorAll("template")
+    );
+
+    return templates.some((template) => {
+      const hasSingleSpaRouter = !!template.content.querySelector(
+        "single-spa-router"
+      );
+      return hasSingleSpaRouter;
+    });
+  })[0];
```
filoxo, line 20: I did this because jest was complaining about Babel's array destructuring helper that was esm and I didn't want to go through the effort of figuring that out.

packages/single-spa-welcome/src/welcome.js
```diff
@@ -1,7 +1,24 @@
...
+    // Search the dom for a single-spa-router element
+    // While this won't work for SSR, the goal of this app is to help
+    // beginners, which likely won't use SSR before running their
+    // first root-config.
+    const templates = Array.prototype.slice.call(
+      document.querySelectorAll("template")
+    );
+
+    return templates.some((template) => {
+      const hasSingleSpaRouter = !!template.content.querySelector(
```
joeldenning, line 15: I don't think that `template.content` is defined in IE11, although this welcome parcel probably is fine without working in IE11.

packages/single-spa-welcome/src/welcome.js
```diff
@@ -1,7 +1,24 @@
...
+import React, { useState } from "react";
 import "./welcome.css";
 
 export default function Root(props) {
+  const isUsingSingleSpaLayout = useState(() => {
+    // Search the dom for a single-spa-router element
+    // While this won't work for SSR, the goal of this app is to help
+    // beginners, which likely won't use SSR before running their
+    // first root-config.
+    const templates = Array.prototype.slice.call(
```
joeldenning, line 10: Maybe easiest to check this way?

```js
return Boolean(document.querySelector('template[id="single-spa-layout"]'))
```

joeldenning (commented):
👍 this will be helpful. A couple small comments.

packages/single-spa-welcome/src/welcome.js
filoxo: Yea I debated whether to do that simple check. Its highly unlikely but still possible that someone changes that id before running their root-config so it would show the wrong instructions. I would just be annoyed if someone opened up an issue caused by that, and I would think a beginner would be frustrated if the instructions don't match up to their code. Maybe I should just create collapsible sections and the user should expand the section with specific instructions themselves?

packages/single-spa-welcome/src/welcome.js
filoxo: Oh poop. Yea, I don't find IE11 to be high priority, but then again, that's why I'm using Array.prototype.slice above instead of Array.from().

packages/single-spa-welcome/src/welcome.js
joeldenning: I'm okay with the longer implementation, too, if you'd prefer it. The longer implementation is more correct for that edge case.

filoxo: Let me think about it. Some of that information can/should be duplicated as part of the ["Getting Started" update](https://github.com/single-spa/single-spa.js.org/issues/335) I suggested and have begun on, and that might help me decide which strategy is best.

joeldenning (approved):

Issue #147, opened 2020-07-09
Update webpack-merge to v5


filoxo:
Update webpack-merge to v5. See https://github.com/survivejs/webpack-merge/commit/534351d5e7dda8f4202cb90bdf6b6a43574de41f

I only took a cursory look at the updates. One change that we are interested in is that webpack-merge changes how to merge configs, specifically that `smart` merging was removed. We use `smart` in the generated Webpack configs to merge single-spa configs with user's configs.

filoxo:
The release notes weren't where I would have expected them to be. This issue seems to be the most comprehensive list of changes: https://github.com/survivejs/webpack-merge/issues/136

filoxo:
After trying to use webpack-merge@5 on a separate project, I'm of the opinion that webpack-merge is no longer useful. Its been completely gutted to the point where its just better not to use it. Here is an example, using the code from their documentation: 

```js
const output = mergeWithCustomize(
  {
    customizeArray(a, b, key) {
      if (key === 'extensions') {
        return _.uniq([...a, ...b]);
      }

      // Fall back to default merging
      return undefined;
    },
    customizeObject(a, b, key) {
      if (key === 'module') {
        // Custom merging
        return _.merge({}, a, b);
      }

      // Fall back to default merging
      return undefined;
    }
  }
)(object1, object2, object3, ...);
```

You can implement the same with just plain lodash (or any other alternative) with just a few functions which I find more readable.

```js
import { mergeWith, uniq, isObject, merge } from 'lodash';

function customize(a, b, key) {
  /* equal to customizeArray */
  if (Array.isArray(a)) { 
    if (key === 'extensions') {
      // custom behavior
      return uniq([...a, ...b]);
    }
    // can explicitly define the default behavior, eg. concat arrays by default
    return a.concat(b)
  }
  /* equal to customizeObject */
  if(isObject(a)){
    if (key === 'module') {
      return mergeWith(a, b, someOtherCustomizeFn);
    }
    // If this function returns `undefined`, merging is handled by the method instead
    // https://lodash.com/docs/4.17.15#mergeWith
    // so this is the same as webpack-merge
    return;
  }
}

const output = mergeWith(object1, object2, object3, ..., customize)
```

I'd be happy to be proven wrong but this is the conclusion I've reached after trying to use the updated webpack-merge for an hour.

joeldenning:
[...]

PR #192, opened 2020-10-07
Feat/generate html template

filoxo:
Nobody asked for it, but I created a generator for single-spa-html to get started with that helper. It's a little rough so I've added some comments.


packages/generator-single-spa/src/generator-single-spa.js
```diff
@@ -128,7 +129,7 @@ async function runFrameworkGenerator() {
         type: "list",
         name: "framework",
         message: "Which framework do you want to use?",
-        choices: ["react", "vue", "angular", "svelte", "other"],
+        choices: ["react", "vue", "angular", "svelte", "html", "other"],
```
filoxo, line 132: I debated calling this `"html/web-component"` but the generated repo doesn't have any web-components code. Should it be left as-is and we can create another generator for web-components? Or should the example app lean into some web-component lib and use that?

packages/generator-single-spa/src/html/templates/src/orgName-projectName.js
```diff
@@ -0,0 +1,14 @@
...
+
+const htmlLifecycles = singleSpaHtml({
+  template,
+});
+
+export const { mount, unmount } = htmlLifecycles;
+
+/* Can also execute JavaScript after application has been mounted
+export const mount = (props) => htmlLifecycles.mount(props)
+  .then(() => { console.log("js after mount"); });
```
filoxo, line 13: Is it worth mentioning this?

packages/generator-single-spa/src/html/templates/webpack.config.js
```diff
@@ -0,0 +1,51 @@
+const { mergeWithRules } = require("webpack-merge");
+const singleSpaDefaults = require("webpack-config-single-spa");
+
+const mergeRulesByTestMatch = mergeWithRules({
```
filoxo, line 4: See this https://github.com/single-spa/create-single-spa/issues/147#issuecomment-705050112 for information.

packages/generator-single-spa/src/generator-single-spa.js
joeldenning: Single-spa-html is not popular at all, from what I've seen. I don't know of anyone actually using it. So it's hard to say what people would expect the generator to create. I think people get excited about the idea of web components and would want some kind of boilerplate. Perhaps a basic lit-html template to help them get started would help, idk.

packages/generator-single-spa/src/html/templates/src/orgName-projectName.js
joeldenning: Since I don't know anyone who actually uses single-spa-html, it's hard to say.

packages/generator-single-spa/src/html/templates/webpack.config.js
joeldenning: Nice - glad to see webpack-merge being more useful again.

filoxo:
I think I'm going to decline this PR for two reasons:

1. we don't have data on single-spa-html usage that will justify adding it to create-single-spa
2. this doesn't provide the starter code for web-components, which could be its own template

This code could perhaps serve as a starting point in the future though.

PR #196, opened 2020-10-11
Adding standalone mode

joeldenning:
See related https://github.com/single-spa/standalone-single-spa-webpack-plugin. Also see https://github.com/react-microfrontends/people which is using the standalone plugin.

### DIFF
--- a/packages/webpack-config-single-spa-react/lib/webpack-config-single-spa-react.js
+++ b/packages/webpack-config-single-spa-react/lib/webpack-config-single-spa-react.js
@@ -3,10 +3,15 @@ const webpackConfigSingleSpa = require("webpack-config-single-spa");
 module.exports = webpackConfigSingleSpaReact;
 
 function webpackConfigSingleSpaReact(opts) {
+  console.log(opts);
+  const webpackConfigEnv = opts.webpackConfigEnv || {};
+
   opts.react = true;
   const config = webpackConfigSingleSpa(opts);
 
-  config.externals.push("react", "react-dom");
+  if (!webpackConfigEnv.standalone) {
+    config.externals.push("react", "react-dom");
+  }
 
   return config;
 }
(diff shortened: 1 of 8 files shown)


joeldenning:
A nice thing of adding this is that now people get an HTML page with instructions when they run a microfrontend and open it up expecting to see their app:

image

packages/webpack-config-single-spa-react/lib/webpack-config-single-spa-react.js
```diff
@@ -3,10 +3,14 @@ const webpackConfigSingleSpa = require("webpack-config-single-spa");
 module.exports = webpackConfigSingleSpaReact;
 
 function webpackConfigSingleSpaReact(opts) {
+  const webpackConfigEnv = opts.webpackConfigEnv || {};
+
   opts.react = true;
   const config = webpackConfigSingleSpa(opts);
 
-  config.externals.push("react", "react-dom");
+  if (!webpackConfigEnv.standalone) {
```
joeldenning, line 11: When in standalone mode, there's no need to use react/react-dom from the import map

filoxo:
The root-config tests are failing with this error: 

```sh
ERROR in Template execution failed: ReferenceError: isLocal is not defined
ERROR in   ReferenceError: isLocal is not defined
```

which seems to be related to how [the root config is merging the the configs](https://github.com/single-spa/create-single-spa/blob/master/packages/generator-single-spa/src/root-config/templates/webpack.config.js#L19-L26).

joeldenning:
ah good catch i'll look into it

joeldenning:
Fixed

filoxo (approved):
🙈  My only remaining question is how will we know whether standalone mode is more helpful than overrides? We don't have any telemetry so it'll be hard to really know when we should take the next step (either go all in, or deprecate the plugin).

joeldenning:
I think that the disabled state of the plugin is quite useful, as it instructs the user to set up import-map-overrides instead of them just seeing a blank screen and being confused.

As more people start using the standalone plugin, I think we'll find out if it ends up being an overall good or bad thing for people trying to understand single-spa.

Issue #109, opened 2020-05-04
Scoped modules with the same name as orgName.


RobertBrewitz:
Good day! Out of the box single-spa uses a regexp to assign all modules scoped with the orgName as external modules.

https://github.com/single-spa/create-single-spa/blob/master/packages/webpack-config-single-spa/lib/webpack-config-single-spa.js#L72

## Problem encountered

What we tried to do before realizing this, was to import npm modules with the same scope name as the orgName.

```javascript
// webpack.config.js
  const defaultConfig = singleSpaDefaults({ orgName: 'my-awesome-org' })

// root.component.tsx
  import * as MyAwesomeOrgPackage from '@my-awesome-org/package';
```

Out of the box this didn't work and webpack treated `@my-awesome-org/package` as an external dependency.

## Hacky solutions

Whilst writing this ticket up we came up with two separate workarounds.

One workaround is to alias `@my-awesome-org/package` without the scope name i.e. `package`.

E.g.

```javascript
// root.component.tsx
  import * as MyAwesomeOrgPackage from 'package';

// tsconfig.json
  "paths": {
    "package": ["@my-awesome-org/package"]
  }

// webpack.config.json
alias: {
  "package": "@my-awesome-org/package",
}
```
Another workaround the we are currently running with is to omit the regexp from the singleSpaDefaults for the specific apps that require the package.

E.g.

```javascript
const ORG_NAME = "my-awesome-org";

module.exports = (webpackConfigEnv) => {
  const defaultConfig = singleSpaDefaults({
    orgName: ORG_NAME,
    projectName: "header",
    webpackConfigEnv,
  });

  defaultConfig.externals = defaultConfig.externals.filter(
    (lib) => !new RegExp(lib).toString().includes(`@${ORG_NAME}`)
  );

  return webpackMerge.smart(defaultConfig, {
    module: {
      rules: [
        {
          test: /\.(s*)css$/,
          use: ["style-loader", "css-loader", "sass-loader"],
        },
      ],
    },
    resolve: {
      extensions: [".js", ".jsx", ".scss"],
    },
  });
};
```

## Request

I am not sure this is an issue really, might just be nice to have some documentation on this topic :)

Kind regards,
Robert

joeldenning:
[...]

PR #199, opened 2020-10-13
add option to not have packages from same org as external

gjhommersom:
See https://github.com/single-spa/create-single-spa/issues/109

Added a setting so that packages from the same org are not marked as external.

Usage: 
```
    // The name of the organization this application is written for
    orgName: 'name-of-company',
    // Boolean flag if all packages that are the same as orgName should be external
    orgPackagesAsExternal: false // default: true
  })
```


joeldenning (approved):
Thanks, this looks good to me. Could you also submit a pull request to the documentation for webpack-config-single-spa? Here's the file to modify:

https://github.com/single-spa/single-spa.js.org/blob/master/website/versioned_docs/version-5.x/create-single-spa.md

PR #200, opened 2020-10-13
Remove trim behavior in validateNaming

### LINKED ISSUE #195: In vue apps, the directory name is not trimmed
Create-single-spa fails if you attempt to create a project with leading or trailing whitespace. We should trim it.

filoxo:
Resolves #195 by removing the string trimming that happens inside of `validateNaming`, which transforms the input for the validation but does not retain that new value. By removing it, space characters will cause a validation error for `orgName` and `projectName`. In the case of the Vue generator, `dir` with spaces will work but then it will prompt for a `projectName` that does not have spaces.


joeldenning (approved):
Looks good to me.

PR #201, opened 2020-10-13
Resolve paths before passing through to commands

### LINKED ISSUE #193: create-single-spa sometimes creates a `.` folder instead of using current working directory
I saw this on someone else's computer, but am not sure how to reproduce it. They were using Node 10 and Windows. Going to try to get more info to diagnose.

filoxo:
Resolves #193

There are only two instances where we use `'.'` as the dir inside of the framework-specific generators. My suspicion is that the one that causes the bug is the Angular one since it's passed as an arg to the angular/cli, while in the Vue one its used to spawn the command with that cwd. In any case, it seems safer to use resolved paths instead of relative ones.


joeldenning (approved):
Nice work, I agree that this will probably fix the issue.

PR #205, opened 2020-10-19
Fix bug in import map poller.

joeldenning:


packages/single-spa-web-server-utils/lib/import-map-poller.js
```diff
@@ -25,7 +25,9 @@ export function getImportMaps({
 }) {
   if (!importMapPromises[url]) {
     importMapPromises[url] = fetchImportMap();
-    setInterval(fetchImportMap, pollInterval);
+    setInterval(() => {
+      importMapPromises[url] = fetchImportMap();
```
joeldenning, line 29: We need to update the importMapPromises object every time we refetch the import map, so that the newly fetched import map is actually used!

filoxo (approved):
Interesting! 🙈

PR #208, opened 2020-10-20
Fix/angular paths again

### LINKED ISSUE #206: Angular directory resolves incorrectly
Hi,

I was using this tool fine last week. but today I am getting this error:
```
C:\repos\microfrontendPoC\app1 >npx create-single-spa --moduleType app-parcel
(node:21216) ExperimentalWarning: The fs.promises API is experimental
? Directory for new project .
? Which framework do you want to use? angular
? Project name (can use letters, numbers, dash or underscore) app1
events.js:174
      throw er; // Unhandled 'error' event
      ^

ReferenceError: path is not defined
    at SingleSpaAngularGenerator.runAngularCli (C:\Users\dev\AppData\Roaming\npm-cache\_npx\21216\node_modules\create-single-spa\node_modules\generator-single-spa\src\angular\generator-single-spa-angular.js:46:17)
    at process._tickCallback (internal/process/next_tick.js:68:7)
Emitted 'error' event at:
    at Immediate.setImmediate (C:\Users\dev\AppData\Roaming\npm-cache\_npx\21216\node_modules\create-single-spa\node_modules\yeoman-generator\lib\index.js:1022:20)
    at runCallback (timers.js:705:18)
    at tryOnImmediate (timers.js:676:5)
    at processImmediate (timers.js:658:5)
```

Any idea what could be happening here?

filoxo:
Resolves #206 by using `dir` as the current working directory instead of the `--directory` path. The angular-cli does not support outputting project files in the current directory so we'll do the same to simplify the logic.

Also, this locks the version of generator-single-spa in create-single-spa so that they're always in lockstep. Otherwise, using an older version (eg. `npx create-single-spa@1.14.1`) will still use the latest minor version of generator-single-spa (eg. `1.17.0`).

### DIFF
--- a/packages/create-single-spa/package.json
+++ b/packages/create-single-spa/package.json
@@ -8,7 +8,7 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "generator-single-spa": "^1.17.1",
+    "generator-single-spa": "1.17.1",
     "yargs": "^14.2.0",
     "yeoman-environment": "^2.6.0"
   }
--- a/packages/generator-single-spa/src/angular/generator-single-spa-angular.js
+++ b/packages/generator-single-spa/src/angular/generator-single-spa-angular.js
@@ -44,18 +44,16 @@ module.exports = class SingleSpaAngularGenerator extends Generator {
       command += ".cmd";
     }
 
-    const cwd = path.resolve(this.options.dir || ".");
+    const cwd = path.resolve(this.options.dir);
 
     const { status, signal } = spawnSync(
       command,
       args.concat([
         "new",
         this.options.projectName, // name of the new workspace and initial project
-        "--directory",
-        cwd,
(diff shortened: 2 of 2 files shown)


packages/create-single-spa/package.json
```diff
@@ -8,7 +8,7 @@
   "author": "",
   "license": "MIT",
   "dependencies": {
-    "generator-single-spa": "^1.17.1",
+    "generator-single-spa": "1.17.1",
```
joeldenning, line 11: Why was this change made?

filoxo: This locks the version of generator-single-spa in create-single-spa so that they're always in lockstep. Otherwise, using an older version (eg. npx create-single-spa@1.14.1) will still use the latest minor version of generator-single-spa (eg. 1.17.0).

joeldenning (approved):

joeldenning (approved):

PR #210, opened 2020-10-23
Disabling HtmlWebpackPlugin and StandaloneSingleSpaPlugin in prod. Resolves #209

### LINKED ISSUE #209: webpack config single-spa breaking existing code possibly due to the standalone plugin
The new changes seems to break the existing projects on upgrade to the new versions of the webpack plugin.

I upgraded my existing single spa app to use the latest versions of single-spa and see there is some issue due to the new `standalone plugin` even when I have not added it to my project

https://github.com/tsukhu/intermix-spa-portal/commit/b13f04069c32f240e49489caa0c3040a09dde782

When I do the build for the sub projects like yarn workspace dashboard build I see a new `index.html` file being added which was I believe specific for to the standalone plugin

image

Also now if you go to https://intermix-spa.netlify.app/ it works but the sub routes if directly accessed (not via the router link) they showing the standalone plugin html file

--- 
Do I need to do anything from my side if this is a breaking change ?

joeldenning:
See #209

### DIFF
--- a/packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
+++ b/packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
@@ -28,6 +28,10 @@ function webpackConfigSingleSpa(opts) {
 
   let webpackConfigEnv = opts.webpackConfigEnv || {};
 
+  let argv = opts.argv || {};
+
+  let isProduction = argv.p || argv.mode === "production";
+
   let HtmlWebpackPlugin = opts.HtmlWebpackPlugin || _HtmlWebpackPlugin;
 
   return {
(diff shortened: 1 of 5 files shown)


packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -28,6 +28,10 @@ function webpackConfigSingleSpa(opts) {
 
   let webpackConfigEnv = opts.webpackConfigEnv || {};
 
+  let argv = opts.argv || {};
```
filoxo, line 31: The Webpack docs seem to lean into using the `env` (or rather `webpackConfigEnv` in our case) parameter to read the production flag rather than directly off of `argv`, as seen here: https://webpack.js.org/configuration/configuration-types/#exporting-a-function. I don't know the difference well enough. Is there a reason to prefer `argv` over `env`?

joeldenning: The webpack env doesn't specify whether it's production mode or not unless you add `--env.prod` to the webpack command. This is a bit redundant, since we already are doing `webpack --mode=production` in the package.json scripts. The `argv` are the arguments passed to webpack's CLI, including the `--mode=production`. Webpack CLI parses the argv with `yargv` in webpack 4 and `commander` in webpack 5.

The reason I chose argv here is to avoid the redundancy of having `webpack --mode=production --env.prod`

filoxo: Ah ok, thanks for explaining! I didn't notice they're different flags.

filoxo (approved):
🙊

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
joeldenning: Yeah the `--mode` flag is one supported by webpack itself, whereas anything starting with `--env` is ignored by webpack but passed onto the webpack configs. In this case, we want to use the `--mode` flag so we get the default webpack behavior for that, but also want the webpack configs to be aware of it.

Issue #198, opened 2020-10-11
Upgrade Webpack configs for Webpack 5


filoxo:
[Webpack 5 has been released](https://webpack.js.org/blog/2020-10-10-webpack-5-release/). I'm opening this issue to 1) begin research into what changes are necessary in webpack-config-single-spa-* packages and if the plugins we use have also been updated, and 2) actually make the changes to be fully compatible with Webpack 5 to publish as a major version change. 

There are a few changes that seem relevant:

> You don’t need to use jsonpFunction anymore.  
>
> PublicPath will be determined automatically. I’m not sure if that means it will work with a systemjs setup or not though 

From: https://single-spa.slack.com/archives/C9JV40F7V/p1602394321086000

adamlusted-swipejobs:
Webpack 5 is required for webpack module federation.

joeldenning:
This is now released in 2.0

filoxo:
Migration guide can be found in the discussion here: https://github.com/single-spa/create-single-spa/discussions/248

PR #216, opened 2020-11-23
Upgrade to webpack 5.

joeldenning:
This is a WIP.

The steps that each project needs to take to upgrade:

1. `yarn add webpack@latest webpack-cli@latest webpack-dev-server@https://github.com/webpack/webpack-dev-server#v4`
2. Change package.json start script from `webpack-dev-server` to `webpack --serve`

We should probably wait for webpack-dev-server@4 before releasing this because we need the fix in https://github.com/webpack/webpack-dev-server/pull/2839, which will be released once https://github.com/webpack/webpack-dev-server/pull/2592 is merged and released.

We should also look into why the devServer.client.port option has to be specified - ideally it wouldn't be.


packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -43,8 +42,9 @@ function webpackConfigSingleSpa(opts) {
       filename: `${opts.orgName}-${opts.projectName}.js`,
       libraryTarget: "system",
       path: path.resolve(process.cwd(), "dist"),
-      jsonpFunction: `webpackJsonp_${opts.projectName}`,
+      // jsonpFunction: `webpackJsonp_${opts.projectName}`,
```
joeldenning, line 46: This option was removed. Not sure if it was renamed to something else. We should look into this more.

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -43,8 +42,9 @@ function webpackConfigSingleSpa(opts) {
...
       devtoolNamespace: `${opts.projectName}`,
+      publicPath: "",
```
joeldenning, line 48: webpack@5 has an "auto" public path option. unfortunately it doesn't work with loading via systemjs right now. Since `"auto"` is the new default for `publicPath`, I had to specify the public path as an empty string so that the public path comes from systemjs-webpack-interop instead of the broken `"auto"` functionality

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -77,14 +77,17 @@ function webpackConfigSingleSpa(opts) {
         },
       ],
     },
-    devtool: "sourcemap",
+    devtool: "source-map",
```
joeldenning, line 81: webpack 5 renamed this

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -77,14 +77,17 @@ function webpackConfigSingleSpa(opts) {
...
     devServer: {
       compress: true,
       historyApiFallback: true,
       headers: {
         "Access-Control-Allow-Origin": "*",
       },
-      disableHostCheck: true,
+      firewall: false,
```
joeldenning, line 88: See https://github.com/webpack/webpack-dev-server/pull/2715

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -77,14 +77,17 @@ function webpackConfigSingleSpa(opts) {
...
+      client: {
+        port: 8080,
```
joeldenning, line 90: `sockPort` was removed - this is the equivalent.

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -94,18 +97,6 @@ function webpackConfigSingleSpa(opts) {
       new BundleAnalyzerPlugin({
         analyzerMode: webpackConfigEnv.analyze ? "server" : "disabled",
       }),
-      new UnusedFilesWebpackPlugin({
```
joeldenning, line 97: This plugin doesn't work with webpack 5 (rip). See https://github.com/tomchentw/unused-files-webpack-plugin/issues/39

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
frehner: https://github.com/single-spa/create-single-spa/issues/198#issue-718889240

frehner: From what I read awhile ago, it’s based on the name field in package.json

sibiraj-s: See https://webpack.js.org/configuration/output/#outputuniquename

joeldenning: Ah thanks I'll switch to that.

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
PaulKujawa: What makes `auto` incompatible with systemJs?

joeldenning: didn't look into it

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
joeldenning: It's much much better not to have to hard code the port here, but let the user specify it via CLI arg - we likely will need to fix webpack-dev-server in order for it to support this. It's one of the things holding up this PR

joeldenning:
The main thing holding up this PR is that it relies on webpack-dev-server@4, which is unreleased

joeldenning:
Ah it looks like webpack-dev-server@4.0.0-beta.0 was just released! I'll try to get this PR ready to merge then - there is still the issue of having to specify port.

joeldenning:
This is now fully updated and the tests should (hopefully) pass - the only thing remaining is the port issue. I plan on making a pull request to webpack-dev-server fixing it.

joeldenning:
The webpack-dev-server issue is here: https://github.com/webpack/webpack-dev-server/issues/2873

joeldenning:
I created a PR to webpack-cli which will resolve the port issue: https://github.com/webpack/webpack-cli/pull/2147

Thumbs up on the PR are appreciated - webpack-cli is a hard project to get code merged into.

joeldenning:
I think we can merge this and publish a beta of create-single-spa with it, noting the caveat when using default port.

I'll mark this as ready for review

joeldenning:
This is released in https://github.com/single-spa/create-single-spa/releases/tag/v2.0.0-beta.0, which includes instructions for upgrading existing projects.

PR #220, opened 2020-11-28
fix(generator-single-spa): add missing dependency chalk and ejs

merceyz:
**What's the problem this PR addresses?**

`generator-single-spa` is trying to use `chalk` and `ejs` without declaring them as dependencies which makes it rely on something else in the dependency tree to provide it and the package manger to hoist it to an accesible location neither of which is guaranteed and under [Yarn PnP](https://yarnpkg.com/features/pnp) is not allowed at all.

https://yarnpkg.com/advanced/rulebook#packages-should-only-ever-require-what-they-formally-list-in-their-dependencies

**How did you fix it?**

Add `chalk` and `ejs` as dependencies

### DIFF
--- a/packages/generator-single-spa/package.json
+++ b/packages/generator-single-spa/package.json
@@ -14,11 +14,12 @@
   "main": "src/generator-single-spa.js",
   "license": "MIT",
   "dependencies": {
+    "chalk": "^3.0.0",
     "command-exists": "^1.2.8",
+    "ejs": "^3.1.5",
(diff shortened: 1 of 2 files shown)


joeldenning (approved):
Thanks - yeah chalk and ejs are dependencies of yeoman, I think, but we should have them here as explicit dependencies.

joeldenning:
Released in https://github.com/single-spa/create-single-spa/releases/tag/v1.18.0

PR #221, opened 2020-11-30
Remove system parser false, as it's the new default in webpack 5.

joeldenning:
I found this when upgrading react-mf/api. Not only is the parser system configuration unnecessary, it actually breaks things in webpack 5 if you try to load json modules.


filoxo (approved):
🙈

Issue #222, opened 2020-11-30
Root project created using 'create-single-spa' is not starting. Failing with ReferenceError: isLocal is not defined


PrakashKumarBalu:
Root project created using 'create-single-spa' is not starting. Failing with ReferenceError: isLocal is not defined

filoxo:
Can you verify whether this line exists in the Webpack config? https://github.com/single-spa/create-single-spa/blob/master/packages/generator-single-spa/src/root-config/templates/webpack.config.js#L29
That is in the current code generated so I'm interested in looking into if that line exists and if not, why.

PrakashKumarBalu:
Yes, this line exists and still failing.

filoxo:
Can you provide better details on how to replicate this? I tried to by:

- Running `npx create-single-spa test-root-config --moduleType root-config`
- Followed the remaining prompts
- Navigated into the test-root-config folder `cd test-root-config`
- And ran `yarn start`

The above steps work correctly and everything compiles as expected. It might also be helpful to include information about your platform and/or node version. Providing a demo repository could also be helpful.

PrakashKumarBalu:
Platform: Windows
Node: v14.3.0

![image](https://user-images.githubusercontent.com/17434336/100698926-ac326800-33d4-11eb-9d4c-e0af13a24f2a.png)

![image](https://user-images.githubusercontent.com/17434336/100698982-c8360980-33d4-11eb-8694-15b3d85c0109.png)

filoxo:
I followed all the steps in the screenshots and am still unable to reproduce. Could you instead share a repo with the code generated on your machine? It could also be platform related, and I don't have a Windows machine to test on at the moment.

filoxo:
[This was also reported on the single-spa Slack by Selom](https://single-spa.slack.com/archives/C8R6U7MT7/p1607348567188700) and they shared a [reproducible example](https://github.com/lomse/debug-single-spa-build-issue). It appears that this is happening because we added the StandaloneSingleSpaPlugin in the base webpack-config-single-spa and that is causing a duplicate HtmlWebpackPlugin to be in the final config.

```sh
[
  CleanWebpackPlugin {
    ...
  },
  BundleAnalyzerPlugin {
    ...
  },
  UnusedFilesWebpackPlugin { ... },
  HtmlWebpackPlugin {        ← first instance to use with StandaloneSingleSpaPlugin that is missing templateParameters
    options: [Object],
    childCompilerHash: undefined,
[...]

PR #228, opened 2020-12-10
Add disableHtmlGeneration option to webpack-config-single-spa

joeldenning:
See #222 and #227


packages/generator-single-spa/src/root-config/templates/package.json
```diff
@@ -37,7 +37,7 @@
     "webpack-cli": "^4.2.0",
     "webpack-config-single-spa": "^1.4.1",
     "webpack-dev-server": "^4.0.0-beta.0",
-    "webpack-merge": "^4.2.2"
+    "webpack-merge": "^5.5.0"
```
joeldenning, line 40: I forgot to upgrade this in my previous PR, so I did so in this PR.

Issue #231, opened 2020-12-13
Solution found: VueJS not working with create-single-spa CLI


tr4g:
**Describe the bug or question**
I just followed the [Playground Guide](http://single-spa-playground.org/playground/) but used `vue` instead of react, and I just keep getting this error all the time, no matter what "solutions" I found on google.

**To Reproduce**
It took me 30 minutes to write all this, so please read it carefully.

### Create a vue project using single-spa CLI:
```sh
mkdir atest ; cd atest
create-single-spa
```
### Pick the following choices:
? Directory for new project **vue-content**
? Select type to generate **single-spa application / parcel**
? Which framework do you want to use? **vue**
? Organization name (can use letters, numbers, dash or underscore) **atest**

Vue CLI v4.5.9
? Please pick a preset:
**❯ Default ([Vue 2] babel, eslint)**
 
 ### Run the project
 ```sh
 cd vue-content
 npm run serve
 ```
 So far so good, you get http://localhost:8080 running just fine, and the CLI added some special single-spa stuff to the `main.js` file:
```
export const bootstrap = vueLifecycles.bootstrap;
export const mount = vueLifecycles.mount;
export const unmount = vueLifecycles.unmount;
```

Now it's time to do get actual single-spa magic, so open another terminal window at `atest` folder and:
 ```sh
 create-single-spa
 ```
### Now pick the following choices:
? Directory for new project **root-config**
? Select type to generate **single-spa root config**
? Which package manager do you want to use? **npm**
? Will this project use Typescript? **No**
? Would you like to use single-spa Layout Engine **No**
? Organization name (can use letters, numbers, dash or underscore) **atest**

### Follow playground instructions
Edit line 47 on `src/index.ejs`to add the vue project
```
      "imports": {
        "@atest/root-config": "//localhost:9000/atest-root-config.js",
        "@atest/vue-content": "//localhost:8080/js/app.js"
      }
```
Add it to the import section (not sure if is needed or not, tried both ways, not mentioned in the guide anyway), line 74 just before the import line already there:
```
    System.import('@atest/vue-content');
```
Now add it it to the `src/atest-root-config.jsp` file, replacing the example welcome code:
```
registerApplication({
  name: "@atest/vue-content",
  app: () =>
    System.import(
      "http://localhost:8080/js/app.js"
    ),
  activeWhen: ["/"],
});
```
### Run the root config project
```sh
cd root-config
npm start
```
[...]

PR #232, opened 2020-12-14
added AMD extra to vue generator, to fix issue

tr4g:
fixed issue #231


frehner:
Thanks! I think we'll move forward with #233 since it updates other dependencies and also removes the unnecessary extra. 

But I want to be clear that I really appreciate you taking the time and effort to do this. 🙏

PR #233, opened 2020-12-14
Fix/missing amd extra

filoxo:
Fixes #233. I also upgraded the other libs while I was at it.


filoxo:
Looks like #232 was also submitted to fix this.

frehner (approved):

PR #235, opened 2020-12-14
Switch to public path as a webpack plugin.

joeldenning:
See 1.x equivalent PR https://github.com/single-spa/create-single-spa/pull/223


filoxo (approved):
🙈

filoxo:
Looks like the tests are failing, possibly because they were triggered before https://github.com/single-spa/create-single-spa/pull/234 got merged. I'll try restarting them.

Issue #237, opened 2020-12-16
single-spa-welcome after init not found


kroissma:
After creating a new single-spa root like explained in the docs and starting via the start script threw the following error:

Uncaught Error: application '@single-spa/welcome' died in status LOADING_SOURCE_CODE: systemjs-webpack-interop: There is no such module '@single-spa/welcome' in the SystemJS registry. Did you misspell the name of your module?

I updated the systemjs-importmap to contain @single-spa/welcome as follow:

```
<script type="systemjs-importmap">
  {
    "imports": {
      "single-spa": "https://cdn.jsdelivr.net/npm/single-spa@5.8.3/lib/system/single-spa.min.js",
      "@single-spa/welcome": "https://unpkg.com/single-spa-welcome/dist/single-spa-welcome.js" // added this line
    }
  }
</script>
```

This resolved my issue. Maybe update the template to fix this issue?

filoxo:
Interesting, thanks for reporting this. I wonder how this is happening since it is included in both templates ([registerApplication version](https://github.com/single-spa/create-single-spa/blob/master/packages/generator-single-spa/src/root-config/templates/src/root-config.ejs#L7) and [single-spa-layout version](https://github.com/single-spa/create-single-spa/blob/master/packages/generator-single-spa/src/root-config/templates/src/layout/index.ejs#L50)). I could see it throwing that error in the single-spa-layout version if built with `prod = true` since that will make the inline import map not render. 

I'm going to try to replicate this on my own but do you think you could provide step-by-step (or even just a screen shot) of all the options that you used? That way it'll save some time.

kroissma:
Here you are:

```
npx create-single-spa --moduleType root-config
? Directory for new project .
? Which package manager do you want to use? npm
? Will this project use Typescript? Yes
? Would you like to use single-spa Layout Engine No
? Organization name (can use letters, numbers, dash or underscore) <project-name>
```

And then just run with `npm start`

joeldenning:
Fixed in https://github.com/single-spa/create-single-spa/releases/tag/v1.18.2. You should be able to just refresh the page and it will automatically download the latest vesrion.

PR #238, opened 2020-12-16
Don't set public path for single-spa-welcome.

joeldenning:
After the switch to using webpack public path as a webpack plugin instead of the set-public-path.js file, it now is looking for a module in the systemjs registry with the name `@single-spa/welcome` which doesn't exist in most import maps.

Since single-spa-welcome isn't using any code splits, I don't think it's necessary to set the public path, so I removed it here.


filoxo (approved):
That makes sense to me.

filoxo:
Would resolve #237?

joeldenning:
Yes this resolves #237

PR #255, opened 2021-01-20
Fixed issue where duplicate versions of the import map overrides were used

PieterBoeren:
Fixed issue where duplicate versions of the import map overrides were used

Upgraded dependencies


joeldenning (approved):
Thanks, this looks good. Will publish a patch in the next day or so with this change.

PR #256, opened 2021-01-21
Improve typescript build times by using write-references.

joeldenning:
From [these docs](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin#typescript-options):

> If you use the babel-loader, it's recommended to use write-references mode to improve initial compilation time.

Since we use babel-loader, I think this will improve build times. I tested locally and it seems to still work as normal.


filoxo (approved):
🙈

PR #259, opened 2021-02-08
Scope single-spa-welcome css.

joeldenning:
See https://github.com/single-spa/single-spa-layout/issues/116 where the global nature of single-spa-welcome's css caused confusion. Generally people shouldn't be using single-spa-welcome at all after initial creation of the root config, but I think it's easy enough to fix this so that others don't run into the same issue.


filoxo (approved):
Yea, this makes sense to do.

PR #266, opened 2021-03-03
Move layout definition to microfrontends-layout.html file.

joeldenning:
The tests for this will fail until https://github.com/single-spa/single-spa-layout/pull/122 is merged and published. This moves the single-spa-layout template for root configs from inside in the index.ejs file to a separate file called microfrontends-layout.html. I think that that file will be easy for people to spot and understand.


packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -71,6 +71,11 @@ function webpackConfigSingleSpa(opts) {
             },
           ],
         },
+        {
+          test: /\.html$/i,
```
joeldenning, line 75: I think we can merge this in without it being a breaking change. It's possible that people have configured `.html` files to be loaded via something other than raw loader and that this causes problems for them when they upgrade. However, I don't think we can account for every possible user-specific change to this shared webpack config when evaluating whether something is breaking.

PR #267, opened 2021-03-04
Make it possible to use a trusted SSL certificate.

joeldenning:
Chrome is getting really aggressive about localhost ssl, having removed the allow-insecure-localhost flag recently and making you type in `thisisinsecure` rather than clicking on "Proceed to localhost"

I'm going to put together documentation for how to get past this, but one step is making it possible to specify the path to your SSL certificates that are trusted by the OS. That's what this PR does.

See related https://github.com/single-spa/single-spa.js.org/pull/420


packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -33,6 +33,45 @@ function webpackConfigSingleSpa(opts) {
 
   let HtmlWebpackPlugin = opts.HtmlWebpackPlugin || _HtmlWebpackPlugin;
 
+  const isHttps = webpackConfigEnv.https;
```
filoxo, line 36: Ok , hear me out. What if none of this is inside of webpack-config-single-spa? It'd be easier for consumers to know what is happening in their config if they're the ones supplying all of this, and leveraging webpack-merge which we already use in every config.

```js
const { merge }  = require("webpack-merge");
const fs = require("fs");
const singleSpaDefaults = require("webpack-config-single-spa");

module.exports = (webpackConfigEnv, argv) => {
    orgName: "example",
    projectName: "app",
    webpackConfigEnv,
    argv,
  });

  return merge(defaultConfig, {
   // modify the webpack config however you'd like to by adding to this object
   host: "localhost",
   https: {
     key: fs.readFileSync('~/.local-ssl/localhost.key'),
     cert: fs.readFileSync('~/.local-ssl/localhost.crt')
   }
  });
};
```

This has the following benefits: 

1. clearer where these certs are coming from
2. avoids needing to use an env var (which then avoids the `npm start --` problem; env vars can be used/read directly in their own config too)
3. keeps the webpack-configs lean
4. still just as extensible

I've had this feeling since `orgPackagesAsExternal` was added because that would have been solved the same way as above (eg. provide your own array of externals to append to/override the internal one). 

However, I do think c-s-s can help here. It could prompt for `"Do you want to use https?"` and if yes, add all the config, prompt if they have existing certs, and setup the rest depending on that. It would add like 4 steps though. Maybe that can be reduced by simply asking if yes/no and then generating boilerplate that is commented out for user configuration after. What do you think about that? Are there any downsides that I've overlooked?

packages/generator-single-spa/src/react/templates/react.package.json
```diff
@@ -2,7 +2,9 @@
   "name": "<%= name %>",
   "scripts": {
     "start": "webpack serve",
+    "start:https": "webpack serve --env https",
```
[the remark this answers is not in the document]
joeldenning: Part of what I'm hoping to accomplish is to document how it's possible to run both http and https. Many organizations end up using both because the root config runs over http but deployed environments are https, so they use http when working on root config and another service, but https otherwise.

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
joeldenning: Here are a few problems I'm trying to solve:

1. Make it possible to avoid insecure warnings without modifying the default generated project, to decrease barrier to entry of using the project. Having to modify the default, generated code is disadvantageous because most people won't know they should, or won't know how to. The TLS problem is very common and not specific to any organization, so I felt it was a good candidate for inclusion in create-single-spa.
2. Provide the documentation and method for avoiding TLS errors in a clear way. Right now one of my consulting clients is typing `thisisinsecure` onto a chrome tab because that's what Chrome requires to visit a site without a secure cert. They don't know how to solve that problem, and I think that create-single-spa and its documentation are a good way of spreading the knowledge.
3. Avoid assumptions about where the certificate is, what format it's in, and what options to provide to webpack-dev-server. In your proposed code, there's an assumption that everyone has the file in the same location and that they generated it with the same method. The `key` and `cert` options are not the only possible options to pass to a TLS server, see https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options. Some file extensions are conventional on windows, where others are conventional on other operating systems. Some people use the `.pem` extension, others don't. Some people use PFX, others don't.

Regarding `orgPackagesAsExternal`, I disagree that it would be better solved by modifying the generated code. Cross-microfrontend imports are a crucial feature of microfrontends, and should be supported out of the box in my opinion.

Maintaining a package like create-single-spa is tricky, as there's always a balance between being opinionated so that the defaults are good, versus not being opinionated enough such that users have to figure out a lot of things on their own. I view localhost TLS certificate errors as a fairly ubiquitous problem, as I've solved it over and over again with every organization I've worked with. I don't see the downsides of merging this PR, as one does not have to use the `SINGLE_SPA_TLS_OPTIONS` environment variable if they don't want to.

joeldenning:
Going to close this, maybe revisit later

PR #276, opened 2021-04-05
Added support for loading images via webpack config

PieterBoeren:
1. Added "asset/resource" for loading images
2. Changed the "raw-loader" for html to "asset/source"
3. Added Typescript types for the images


PieterBoeren:
Tests seems to fail (seems not really related to this change), can't get them to work on my machine either...

packages/generator-single-spa/src/common-templates/typescript/declarations.d.ts
```diff
@@ -2,3 +2,44 @@ declare module "*.html" {
...
+
+declare module "*.webp" {
+  const src: string;
+  export default src;
+}
+
+declare module "*.svg" {
+  import * as React from "react";
+
+  export const ReactComponent: React.FunctionComponent<
```
joeldenning, line 39: Will this throw errors in utility modules that don't have react installed? we can change declarations.d.ts to be declarations.d.ts.ejs and then use `<% if (react) { $>` to include this in react projects but not in other projects.

joeldenning (commented):
Looks good, one comment. Nice catch with removing raw-loader

joeldenning:
The tests usually fail due to race conditions with `pnpm installs`. I'm working on a fix for them

packages/generator-single-spa/src/common-templates/typescript/declarations.d.ts
PieterBoeren: Check, I'm so used to React, that I didn't even spotted this one. Fixed as proposed.

packages/generator-single-spa/src/common-templates/typescript/declarations.d.ts.ejs
```diff
@@ -0,0 +1,47 @@
...
+declare module "*.webp" {
+  const src: string;
+  export default src;
+}
+
+<% if (react) { $>
+declare module "*.svg" {
+  import * as React from "react";
+
+  export const ReactComponent: React.FunctionComponent<
```
joeldenning, line 40: As I think about this more, I'm confused. We're not using react-svg-loader, so why would we need to export a ReactComponent from svg files?

packages/generator-single-spa/src/common-templates/typescript/declarations.d.ts.ejs
```diff
@@ -0,0 +1,47 @@
...
+  const src: string;
+  export default src;
+}
+
+declare module "*.webp" {
+  const src: string;
+  export default src;
+}
+
+<% if (react) { $>
```
joeldenning, line 36: svgs should be supported in non-react projects. Either through an `<% else %>` or by removing the (possibly unnecessary) react references inside of the svg.

packages/generator-single-spa/src/common-templates/typescript/declarations.d.ts.ejs
PieterBoeren: You're not confused, I was :) I took my inspiration from the create-react-app and didn't consider the non-react scenario at all. I will look into this tomorrow...

joeldenning: 👍🏿  sounds good. After this is addressed I will merge and publish

PieterBoeren: This is getting a bit complex. When declaring a simple type for the svg files (just like all the other image types, just a "src"), things do work. However, I guess that React developers expect the above declaration so they could use the svg both as simple image source, as well as true React component (see: https://create-react-app.dev/docs/adding-images-fonts-and-files#adding-svgs). For that we would need the above declaration (but with the if check inside the type, to only include the React stuff when targeting React).

However, in order to be able to support this, the webpack config also has to be changed (with: https://react-svgr.com/docs/webpack/). So I need some more time to figure out the correct configuration :)

PieterBoeren: I removed the React SVG component, couldn't get it to work properly.

To be clear things like: 

```
import logoSvg from "./logo.svg"

<img src={logoSvg} />
```

do work with this change.

joeldenning (approved):
Looks good

joeldenning:
I should not have merged this - it has several issues and breaks existing functionality without adding the new functionality, since the declarations file was renamed but the reference to it wasn't. Will follow up with a PR to fix

PR #281, opened 2021-04-09
Fix start instructions for pnpm. Resolves #278.

### LINKED ISSUE #278: Project startup instructions when using pnpm are incorrect
When creating a new single-spa application with npmn (I used Typescript, but that shouldn't matter) the instructions to start the project are not working. It states "Run 'pnpm start --port 8500'", but this gives an error: 

```"Unknown option: 'port'".```

I guess it is not passing the option to the start script, I worked around it by adding the port directly inside the start script ("start": "webpack serve --port 8500").

joeldenning:
See #278. pnpm and npm require the extra `--` in the command. When pnpm was added as a new option, this was overlooked.

### DIFF
--- a/.changeset/stupid-mayflies-visit.md
+++ b/.changeset/stupid-mayflies-visit.md
@@ -1,0 +1,5 @@
+---
+"generator-single-spa": patch
+---
+
+Fix start instructions for pnpm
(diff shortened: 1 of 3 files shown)


joeldenning:
Tests are failing due to race conditions with concurrent pnpm installs. Not related to this code

TheMcMurder (approved):

PR #284, opened 2021-04-22
fix: vscode warning

pmjhonwang:
CleanShot 2021-04-22 at 17 16 01@2x


joeldenning:
Thanks for your contribution.  Please provide a changeset with this change - here are the instructions: https://github.com/single-spa/create-single-spa/blob/main/CONTRIBUTING.md#add-changeset-for-each-pr. After you do that the build will get closer to succeeding but will still probably fail due to some race conditions with pnpm installs. However, those are not new so we'll be able to merge this once it has a changeset.

joeldenning:
Closing due to inactivity

PR #288, opened 2021-04-28
fix: child application async chunk path conflict

Janlay884181317:
In Webpack4, when the child application both use async route chunk( import() ), and if a file path is the same (for example, there is a utils/index.js) but the export content is different, an error will be reported.

But in webpack5, `output.jsonpFunction` value get from `package.json` => `name` field
https://webpack.js.org/blog/2020-10-10-webpack-5-release/#automatic-unique-naming

### DIFF
--- a/packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
+++ b/packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
@@ -46,6 +46,7 @@ function webpackConfigSingleSpa(opts) {
       uniqueName: opts.projectName,
       devtoolNamespace: `${opts.projectName}`,
       publicPath: "",
+      jsonpFunction = `webpackJsonp_${opts.projectName}`;
     },
     module: {
       rules: [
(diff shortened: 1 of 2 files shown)


packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -46,6 +46,7 @@ function webpackConfigSingleSpa(opts) {
       uniqueName: opts.projectName,
       devtoolNamespace: `${opts.projectName}`,
       publicPath: "",
+      jsonpFunction: `webpackJsonp_${opts.projectName}`;
```
joeldenning, line 49: `jsonpFunction` was removed in webpack 5, and create-single-spa now uses webpack 5. If you're wanting to use webpack 4, use create-single-spa@1.

See https://webpack.js.org/blog/2020-10-10-webpack-5-release/#automatic-unique-naming for the explanation.

joeldenning (commented):
create-single-spa@2 only supports webpack 5, but this change seems to be for webpack 4.

PR #290, opened 2021-04-29
Fix semver check. Resolves #285.

### LINKED ISSUE #285: Patch update message is incorrect
Minor issue, but when running "create-single-spa" command, I get this message:

```
"A patch update of create-single-spa is available: 2.3.1 → 2.2.2"
```

It seems that the upgrade check is only taking the patch version into consideration and not the entire version.

joeldenning:
See #285

### DIFF
--- a/.changeset/wicked-years-tan.md
+++ b/.changeset/wicked-years-tan.md
@@ -1,0 +1,5 @@
+---
+"generator-single-spa": patch
+---
+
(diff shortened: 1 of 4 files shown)


joeldenning:
Tests will fail because they're generally broken due to race conditions. But the tests specific to this PR pass locally.

PR #292, opened 2021-05-06
fixed generated jest config for react unable to load parcels 

gjhommersom:
Changes made due to issues found here: https://github.com/single-spa/single-spa-react/issues/103

Will not fix existing projects but newly generated projects should work ok now. 
Existing projects can be manually fixed by using the example code @joeldenning provided.

### DIFF
--- a/.changeset/seven-windows-attend.md
+++ b/.changeset/seven-windows-attend.md
@@ -1,0 +1,5 @@
+---
+"generator-single-spa": patch
+---
+
+fixed generated jest config for react unable to load parcels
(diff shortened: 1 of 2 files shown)


joeldenning (approved):
This looks good. I pulled down the changes locally to test them, and verified that with this change jest is able to properly load the parcel component where it previously could not.

joeldenning:
Released in https://github.com/single-spa/create-single-spa/blob/main/packages/generator-single-spa/CHANGELOG.md#300

PR #294, opened 2021-05-09
fix: noscript tag

cejaramillof:
Changes made due to issues found here: https://github.com/single-spa/create-single-spa/issues/213

add NoScript tag to HTML templates

### DIFF
--- a/.changeset/nine-apes-admire.md
+++ b/.changeset/nine-apes-admire.md
@@ -1,0 +1,5 @@
+---
+"generator-single-spa": patch
+---
+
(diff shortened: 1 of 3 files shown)


joeldenning (approved):
Thanks, it's a sensible addition.

joeldenning:
Build is currently failing on main branch, so it's not caused by this change. Merging

joeldenning:
released in https://github.com/single-spa/create-single-spa/blob/main/packages/generator-single-spa/CHANGELOG.md#300

PR #298, opened 2021-06-25
Migrating the boilerplate to the new JSX Transform and minor changes in package.json and webpack.

EduSantosBrito:


packages/generator-single-spa/src/root-config/templates/root-config.package.json
```diff
@@ -25,6 +25,7 @@
     "eslint-config-important-stuff": "^1.1.0",
     "eslint-config-prettier": "^7.0.0",
     "eslint-plugin-prettier": "^3.2.0",
+    "html-webpack-plugin": "^5.3.1",
```
joeldenning, line 28: Good catch! html-webpack-plugin is a dependency of webpack-config-single-spa but we're turning off html generation from webpack-config-single-spa and requiring html-webpack-plugin directly inside of the root config webpack config. So this should be a dependency of root configs 👍🏿

packages/generator-single-spa/src/common-templates/babel.config.json.ejs
```diff
@@ -1,7 +1,12 @@
 {
   "presets": [
     "@babel/preset-env",<% if (framework === 'react') { %>
-    "@babel/preset-react",<% } if (typescript) { %>
+    [
+      "@babel/preset-react", 
+      {
+        runtime: "automatic"
```
joeldenning, line 7: 👍🏿  I agree this is a good change

joeldenning (approved):

joeldenning:
Thanks for the changes - they're all good

joeldenning:
The tests have been flaky for a long time, but I had tonight to fix them. In doing so, it revealed that these changes caused typescript problems, since the tsconfig needs to know to use the `_jsx` function from react when it is type checking. I fixed it (and the test flakiness) in https://github.com/single-spa/create-single-spa/pull/299/files#diff-b109d4cb8b90147bfc558c880e68451ba053c8d744d3cb32553d531459612a32R4

I'll merge that change and a couple others and then publish.

joeldenning:
Released in https://github.com/single-spa/create-single-spa/blob/main/packages/generator-single-spa/CHANGELOG.md#300

PR #303, opened 2021-07-01
npm install fails with svelte due to missing husky dependency

cbryer:


packages/generator-single-spa/src/svelte/templates/svelte.package.json
```diff
@@ -17,6 +17,7 @@
     "@testing-library/jest-dom": "^5.14.1",
     "@testing-library/svelte": "^3.0.3",
     "babel-jest": "^27.0.5",
+    "husky": "^6.0.0",
```
joeldenning, line 20: Could you increase this to husky 7?

joeldenning (changes requested):
Thanks for the contribution - could you add a changeset, as described in the comment above and in the CONTRIBUTING.md?

joeldenning:
Closing due to inactivity

Issue #309, opened 2021-07-20
Dev server failing


alexristich:
In initializing a new project with `create-single-spa`, I ran into an issue with the webpack server:

```
[webpack-cli] Invalid options object. Dev Server has been initialized using an options object that does not match the API schema.
 - options has an unknown property 'firewall'. These properties are valid:
   object { allowedHosts?, bonjour?, client?, compress?, devMiddleware?, headers?, historyApiFallback?, host?, hot?, http2?, https?, ipc?, liveReload?, onAfterSetupMiddleware?, onBeforeSetupMiddleware?, onListening?, open?, port?, proxy?, setupExitSignals?, static?, watchFiles?, webSocketServer? }
 - options.client has an unknown property 'host'. These properties are valid:
   object { logging?, overlay?, progress?, webSocketTransport?, webSocketURL? }
```

In taking a closer look, it seems `webpack-dev-server`, a beta release, put out their [first release candidate](
https://github.com/webpack/webpack-dev-server/releases/tag/v4.0.0-rc.0) yesterday. This introduced some breaking changes, hence the incompatibility.

I was able to get this working locally by modifying `packages/webpack-config-single-spa/lib/webpack-config-single-spa.js` to reflect the recent changes. I'll open a PR shortly with the proposed changes.

joeldenning:
Fix released in webpack-config-single-spa@4.0.0. See https://github.com/single-spa/create-single-spa/blob/main/packages/webpack-config-single-spa/CHANGELOG.md#400 for migration instructions

alexristich:
Hey @joeldenning, seems this doesn't yet work with `yarn create single-spa --framework react`. I'm struggling to follow exactly where the dependencies lie, but it seems `create-single-spa` uses `generator-single-spa`, and in https://github.com/single-spa/create-single-spa/commit/5dc82a6ce97a72a53dc2533fe45d2f02504be4e9 there weren't any changes in `generator-single-spa`'s package.json file to include the new versions of `webpack-config-single-spa{-react,-ts,-react-ts}`, so it's still using `webpack-config-single-spa@3.1.0` in the dependencies:

```
webpack-config-single-spa-ts@2.2.4, webpack-config-single-spa-ts@^2.2.2:
  version "2.2.4"
[...]

PR #310, opened 2021-07-20
Upgrade to latest config from webpack-dev-server v4.0.0-rc.0

alexristich:
Resolves https://github.com/single-spa/create-single-spa/issues/309

I'm guessing this would be considered a major release given it'd break for anyone who might still be using `webpack-dev-server` v3.x.x (or anyone who's pinned to a beta release of `webpack-dev-server` v4.0.0-beta(x).

Happy to make any changes, just thought I'd kick off a PR to hopefully save some time in getting this fixed :)

### DIFF
--- a/packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
+++ b/packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
@@ -89,6 +89,11 @@ function webpackConfigSingleSpa(opts) {
       headers: {
         "Access-Control-Allow-Origin": "*",
       },
+      client: {
+        webSocketURL: {
+          hostname: "localhost",
+        },
+      },
       allowedHosts: "all",
     },
     externals: opts.orgPackagesAsExternal
--- a/packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
+++ b/packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
@@ -85,15 +85,11 @@ function webpackConfigSingleSpa(opts) {
     },
     devtool: "source-map",
     devServer: {
-      compress: true,
       historyApiFallback: true,
       headers: {
         "Access-Control-Allow-Origin": "*",
       },
-      firewall: false,
-      client: {
-        host: "localhost",
-      },
+      allowedHosts: "all",
     },
     externals: opts.orgPackagesAsExternal
       ? ["single-spa", new RegExp(`^@${opts.orgName}/`)]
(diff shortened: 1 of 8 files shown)


veronesecoms (approved):

joeldenning (approved):
Thanks for your help in reporting the issue, identifying the cause, and contributing a fix. I've pushed a couple commits to this PR to fix up the following:

- The version of webpack-dev-server in newly generated package.jsons should be 4.0.0-rc.0
- The client.host property was renamed to client.webSocketURL.hostname - it was not removed. It's important because otherwise the web socket connectino from the browser to webpack-dev-server will not work since it will make the request relative to the html page (root config) rather than the microfrontend's webpack-dev-server.

I also agree this is a breaking change and will publish it as a new major for webpack-config-single-spa

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -85,15 +85,16 @@ function webpackConfigSingleSpa(opts) {
     },
     devtool: "source-map",
     devServer: {
-      compress: true,
```
joeldenning, line 88: I confirmed that this is the new default value in webpack-dev-server and that gzip compression still occurs with this change.

joeldenning:
Fix released in webpack-config-single-spa@4.0.0. See https://github.com/single-spa/create-single-spa/blob/main/packages/webpack-config-single-spa/CHANGELOG.md#400 for migration instructions

alexristich:
Thanks for tidying this up! Haven't worked with `pnpm` before so I was reluctant to commit the `pnpm-lock.yaml` file in the first commit. I see now I had no reason to worry. :)

> The client.host property was renamed to client.webSocketURL.hostname - it was not removed. It's important because otherwise the web socket connectino from the browser to webpack-dev-server will not work since it will make the request relative to the html page (root config) rather than the microfrontend's webpack-dev-server.

Ah, of course. I couldn't figure out where it went and thought this might have been superseded by [`host`](https://webpack.js.org/configuration/dev-server/#devserverhost) which has a default value of `localhost`. Your explanation makes way more sense - thanks!

joeldenning:
> I couldn't figure out where it went and thought this might have been superseded by host which has a default value of localhost

Yeah I wasn't sure if `host`'s default value would be enough, but in local testing it was not, unfortunately. I think ideally we wouldn't need to specify client.webSocketURL.hostname, but it seems required for now.

PR #316, opened 2021-08-16
Output typescript declarations during build. Resolves #289.

### LINKED ISSUE #289: Output declarations by default on utility modules
On utility module making, there should be templated so that the declarations are created by default.
Right now there is no documentation or simple way to get declartions. One have to set it up self.
This issue was raised: https://github.com/single-spa/single-spa.js.org/issues/361#issuecomment-827027568

And it was agreed that declarations should be on by default for utility modules.

Right now I dont have time to add it, as Im raising forward to get our rebuild up and running and ready for release. If I find the time I will come back and fix it.

But the issue is here for a tracker.

joeldenning:
See #289

### DIFF
--- a/packages/generator-single-spa/src/common-templates/typescript/typescript.package.json
+++ b/packages/generator-single-spa/src/common-templates/typescript/typescript.package.json
@@ -1,4 +1,7 @@
 {
+  "scripts": {
+    "build:types": "tsc"
+  },
   "devDependencies": {
     "@babel/preset-typescript": "^7.14.5",
     "eslint-config-ts-important-stuff": "^1.1.0",
@@ -9,6 +12,7 @@
   "dependencies": {
     "@types/jest": "^26.0.23",
     "@types/systemjs": "^6.1.0",
-    "@types/webpack-env": "^1.16.0"
+    "@types/webpack-env": "^1.16.0",
+    "single-spa": "^5.9.3"
   }
 }
--- a/packages/generator-single-spa/src/react/generator-single-spa-react.js
+++ b/packages/generator-single-spa/src/react/generator-single-spa-react.js
@@ -58,6 +58,9 @@ module.exports = class SingleSpaReactGenerator extends PnpmGenerator {
     Object.assign(this.options, answers, { framework: "react" });
   }
   async createPackageJson() {
+    this.srcFileExtension = this.options.typescript ? "tsx" : "js";
+    this.mainFile = `src/${this.options.orgName}-${this.options.projectName}.${this.srcFileExtension}`;
+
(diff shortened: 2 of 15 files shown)


packages/generator-single-spa/src/common-templates/typescript/typescript.package.json
```diff
@@ -9,6 +12,7 @@
   "dependencies": {
     "@types/jest": "^26.0.23",
     "@types/systemjs": "^6.1.0",
-    "@types/webpack-env": "^1.16.0"
+    "@types/webpack-env": "^1.16.0",
+    "single-spa": "^5.9.3"
```
joeldenning, line 16: This single-spa-react types rely on the single-spa types, so I added this dependency. This file is used by more than just react applications, but I figure having the single-spa library and its types available is not a bad idea for util modules / root configs, as well

packages/generator-single-spa/src/react/generator-single-spa-react.js
```diff
@@ -79,6 +83,8 @@ module.exports = class SingleSpaReactGenerator extends PnpmGenerator {
       delete packageJson.devDependencies["webpack-config-single-spa"];
       // Will be replaced by webpack-config-single-spa-react-ts
       delete packageJson.devDependencies["webpack-config-single-spa-ts"];
+
+      packageJson.types = `dist/${this.options.orgName}-${this.options.projectName}.d.ts`;
```
joeldenning, line 87: In case people publish the packages to npm, the `"types"` field should be properly set

packages/generator-single-spa/src/react/templates/tsconfig.json
```diff
@@ -1,8 +1,10 @@
 {
   "extends": "ts-config-single-spa",
   "compilerOptions": {
-    "jsx": "react-jsx"
+    "jsx": "react-jsx",
+    "declarationDir": "dist"
   },
-  "include": ["src/**/*", "node_modules/@types"],
+  "files": ["<%- mainFile %>"],
```
joeldenning, line 7: The `"files"` configuration is necessary for typescript to know which file is the main file.

packages/ts-config-single-spa/tsconfig.json
```diff
@@ -14,6 +14,7 @@
       "es2016.array.include",
       "es2018"
     ],
-    "noEmit": true
+    "declaration": true,
```
joeldenning, line 17: These changes are what makes this a breaking change - previously no types emitted, now types are emitted

TheMcMurder (approved):
Looks good to me

PR #317, opened 2021-08-16
Support for React utility microfrontends. Resolves #264

### LINKED ISSUE #264: React-based utility modules
Perhaps `create-single-spa` should ask if a new utility module needs framework support (in our case React) so that the webpack config will be generated with the correct package dependency.

We had some React components that were being re-used heavily, so we decided to put them into a utility module and pre-load it from the root-config. I ran `create-single-spa` and got everything set up, moved the components over and exported them, and to my dismay React started crashing with the misleading [Invalid Hook Call](https://reactjs.org/warnings/invalid-hook-call-warning.html) error. That led to [this](https://github.com/facebook/react/issues/13991) enormous, years-long thread-of-misery, and all I could tell for sure is that multiple _copies_ of React were being loaded (complicated by the fact that most people seem to be fighting multiple _versions_ of React, which isn't possible in our hoisted monorepo).

Anyway, as I'm sure you've guessed, the problem is that the utility's webpack config was generated by `webpack-config-single-spa`, and the fix was to instead require `webpack-config-single-spa-react` which declares React as a webpack external.

Considering (a) the error message from React is terrible and misleading (it has nothing to do with hooks...), (b) troubleshooting the problem is difficult at best, and (c) I only recognized the solution (after several hours) because I'd previously dug into the single-spa config packages -- this seems like something people could lose a lot of time chasing, which might be easily avoided with a simple change to the script.

joeldenning:
See #264

### DIFF
--- a/tests/test-helpers.js
+++ b/tests/test-helpers.js
@@ -14,7 +14,7 @@ exports.createFixtureIfDoesntExist = function (name, args) {
     return new Promise((resolve, reject) => {
       const argsStr = args.replace(/\s+/g, " ");
 
-      console.log(`Creating '${name}' fixture. This could take a while...`);
+      console.log(`Creating '${name}' fixture.`);
 
       const command = `node ../../packages/create-single-spa/bin/create-single-spa.js --dir ${name} ${argsStr}`;
 
--- a/.changeset/odd-paws-wink.md
+++ b/.changeset/odd-paws-wink.md
@@ -1,0 +1,12 @@
+---
+"generator-single-spa": major
+"create-single-spa": major
+---
+
+Breaking changes:
+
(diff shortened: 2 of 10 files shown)


tests/test-helpers.js
```diff
@@ -14,7 +14,7 @@ exports.createFixtureIfDoesntExist = function (name, args) {
     return new Promise((resolve, reject) => {
       const argsStr = args.replace(/\s+/g, " ");
 
-      console.log(`Creating '${name}' fixture. This could take a while...`);
```
joeldenning, line 17: The tests got faster once we added `--skipInstall` to them.

TheMcMurder (approved):
🐙

PR #319, opened 2021-08-18
Added missing quotes in the babel config

PieterBoeren:
I noticed some missing quotes in the generated babel config, added those.

### DIFF
--- a/.changeset/wise-birds-lick.md
+++ b/.changeset/wise-birds-lick.md
@@ -1,5 +1,5 @@
 ---
-"generator-single-spa": minor
+"generator-single-spa": patch
 ---
 
 Fixed small issue in babel config file
--- a/.changeset/wise-birds-lick.md
+++ b/.changeset/wise-birds-lick.md
@@ -1,0 +1,5 @@
+---
+"generator-single-spa": minor
+---
+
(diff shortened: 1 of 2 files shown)


.changeset/wise-birds-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+"generator-single-spa": minor
```
joeldenning, line 2: This should be a patch, not a minor

joeldenning (commented):
Babel config files are not strict json, so the quotes aren't necessary. Is there something the quotes help with? Maybe with syntax highlighting in VS Code, which might not realize that babel configs can be JSON 5?

PieterBoeren:
It was indeed VS Code which made it quite red, but also the rest of the options in the file are with quotes, so for me, it's more consistent. Let me know if you agree, then I will fix the changes tomorrow.

joeldenning:
Sure, we can release this as a patch - please address the comment above and then I'll approve and merge.

joeldenning:
Released in generator-single-spa@4.0.2. See https://github.com/single-spa/create-single-spa/blob/main/packages/generator-single-spa/CHANGELOG.md#402

PR #335, opened 2021-09-20
Remove dependency on unused-files-webpack-plugin. Resolves #334.

### LINKED ISSUE #334: unused-files-webpack-plugin isn't used anymore, but still is a dependency
### Discussed in https://github.com/single-spa/create-single-spa/discussions/331

<div type='discussions-op-text'>

<sup>Originally posted by **SergioTx** September 14, 2021</sup>
I have seen it as a project dependency, but it is used nowhere in the project. I have checked other dependencies and still nothing.

This comes because I have an issue with that plugin that reports all the files that only contain TS interfaces as unused.

Is there a way to disable it in certain files? Or to disable it completely?</div>

joeldenning:
See #334 and #331

### DIFF
--- a/.changeset/loud-knives-tap.md
+++ b/.changeset/loud-knives-tap.md
@@ -1,0 +1,5 @@
+---
+"webpack-config-single-spa": patch
+---
+
+Remove dependency on unused-files-webpack-plugin. Resolves #334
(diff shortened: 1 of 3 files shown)


TheMcMurder (approved):

joeldenning:
This is released in webpack-config-single-spa@5.1.1. See https://github.com/single-spa/create-single-spa/blob/main/packages/webpack-config-single-spa/CHANGELOG.md#511

PR #344, opened 2021-11-30
fix upgrade-dependencies script

lipsumar:
This PR changes the name `ncu` to `npm-check-updates`. While `ncu` is the name of the binary exported by `npm-check-updates`, `ncu` is actually [another package](https://www.npmjs.com/package/ncu).

After this fix, i was able to run the script locally without having `npm-check-updates` installed globally.


lipsumar:
I believe this change does not impact package versions, so having no changeset should be fine. Please advise if I got this wrong.

joeldenning (approved):
Thanks for this correction

PR #364, opened 2022-10-17
fix: upgrading fork-ts-checker-webpack-plugin version and configuration

anthonysessa:
When linking a package, e.g. yalc, yarn link, etc., the [fork-ts-checker-webpack-pluing](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin) does not rebuild TypeScript, so webpack-dev-server cannot find updated exports.

According to the [documentation](https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/#options), when using `mode: "write-references"` the `build` flag should be set to true.

>Use readonly if you don't want to write anything on the disk, write-dts to write only .d.ts files, write-tsbuildinfo to write only .tsbuildinfo files, write-references to write both .js and .d.ts files of project references (last 2 modes requires build: true).

Also updated `fork-ts-checker-webpack-plugin` to the latest release.

### DIFF
--- a/packages/webpack-config-single-spa-ts/lib/webpack-config-single-spa-ts.js
+++ b/packages/webpack-config-single-spa-ts/lib/webpack-config-single-spa-ts.js
@@ -23,6 +23,7 @@ const modifyConfig = (opts, webpackConfig) => {
             typescript: {
               mode: "write-references",
             },
+            build: true,
           }),
     ].filter(Boolean),
     resolve: {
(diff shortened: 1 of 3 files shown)


anthonysessa:
Wanted to ping @joeldenning  @filoxo @TheMcMurder to get a review when possible

joeldenning (commented):
Looks good to me, I don't do much maintenance anymore so I won't be able to come back to merge and publish it, but maybe the other maintainers can. There's a conflict with pnpm-lock.yaml that should be resolved (start by taking the one from main branch and discarding your changes)

PR #376, opened 2023-04-08
ForkTsCheckerWebpackPlugin requires build:true

stoopkidddd:
According to the documentation for fork-ts-checker-webpack-plugin, when you set `mode: 'write-references`, it is required to also set `build: true`.

See here - https://github.com/TypeStrong/fork-ts-checker-webpack-plugin#typescript-options

I encountered when working in a monorepo with a single-spa application that imports a single-spa utility. I realized that if you are running webpack locally for both, the TS checker does not recompile and you will get incorrect/out of date errors when you change the contract of a function.

### DIFF
--- a/.changeset/nine-trains-do.md
+++ b/.changeset/nine-trains-do.md
@@ -1,0 +1,5 @@
+---
+"webpack-config-single-spa-ts": minor
+---
+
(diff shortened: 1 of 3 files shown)


anthonysessa:
I had this commit open that was meant to solve this issue but it is now out of sync.  I am going to close it: https://github.com/single-spa/create-single-spa/pull/364

cc @filoxo

Issue #377, opened 2023-05-08
Error typescript version dependency 


guilhermegregio:
## Reproduce

Install a new version of typescript and install package `webpack-config-single-spa-ts`

## Error  

error An unexpected error occurred: "could not find a copy of typescript to link in /app/node_modules/webpack-config-single-spa-ts/node_modules".

guilhermegregio:
Posible solution

Remove typescript of dependencies from package `webpack-config-single-spa-ts`

adripanico:
As @guilhermegregio states, I don't think neither `webpack-config-single-spa-ts` nor `webpack-config-single-spa-react-ts` actually require having TypeScript as dependency. For the moment, we have decided to add both configurations to our project, removing such dependencies and adding `fork-ts-checker-webpack-plugin` and `webpack-config-single-spa`. This way we have been able to properly bump TS to latest version and everything is working as expected.

Gedzis:
Hello,
any updates on this issue?

ValchanOficial:
Same problem here =/

MilanKovacic:
I will take a look at this over the weekend.

ludeviance:
Same here with typescipt@5.x.x as a devDependency on my project. If I downgrade to typescript 4 or lower, the error is gone, but my project can't use lower than 5.

MilanKovacic:
I was unable to reproduce the issue by creating a new root config / application, and upgrading to the latest typescript 5 version.
I suspect the problem is the declaration of typescript in both dependencies, and peer dependencies in the webpack-config-single-spa-ts package.

Can you please specify which package manager you are using, and post a small reproduction repository?

Gedzis:
Hello,
I tried to reproduce and still exists. Steps to reproduce:

1) Create new `single-spa root config` project with `yarn create single-spa`
2) Update dependencies to:
* "single-spa": "^6.0.2",
* "single-spa-layout": "^2.2.0"
3) Run `yarn install`
4) Upgrade typescript dependency to "typescript": "^5.6.2",
5) Run `yarn install` and error is `error Invariant Violation: could not find a copy of typescript to link in`

joeldenning:
I am able to replicate the issue following those steps. Note that pnpm doesn't seem to have the same issue, so perhaps it is yarn-specific.

Taifunov:
Up

felipe-gustavo:
I got this recently, add the specific typescript version on `package.json` solves it.

[...]

PR #378, opened 2023-05-08
fix typescript version error

guilhermegregio:
## Solve error:

error An unexpected error occurred: "could not find a copy of typescript to link in /app/node_modules/webpack-config-single-spa-ts/node_modules".

Issue: #377


TheMcMurder:
Merging this created some build errors. I've reverted it for now while I look into the errors and try to dust off my brain on what's happening.

PR #379, opened 2023-05-11
fix typescript version error

guilhermegregio:
## Solve error:

error An unexpected error occurred: "could not find a copy of typescript to link in /app/node_modules/webpack-config-single-spa-ts/node_modules".

Issue: #377

### DIFF
--- a/.changeset/spotty-tigers-tap.md
+++ b/.changeset/spotty-tigers-tap.md
@@ -1,0 +1,5 @@
+---
+"webpack-config-single-spa-ts": patch
+---
+
+fix typescript version error
--- a/packages/webpack-config-single-spa-ts/package.json
+++ b/packages/webpack-config-single-spa-ts/package.json
@@ -30,9 +30,11 @@
   },
   "dependencies": {
     "fork-ts-checker-webpack-plugin": "^6.3.2",
-    "typescript": "^4.1.2",
     "webpack-config-single-spa": "workspace:*",
     "webpack-merge": "^5.8.0"
+  },
+  "devDependencies": {
+    "typescript": "^4.1.2"
   },
   "peerDependencies": {
     "typescript": ">=4"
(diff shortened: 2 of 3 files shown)


guilhermegregio:
solved: error of builds @TheMcMurder, i saw i do not run tests locale and not create changes on lock file causing the build errors, verify again please

adripanico:
Change looks good. Could someone take a look to it, please? @joeldenning, maybe? Thanks

joeldenning:
I tested this out locally and it didn't seem to fix the issue. However, I'm not sure because create-single-spa uses pnpm and the issue my be yarn-specific. I tried yarn linking the create-single-spa package, which might not be the same as if it were published to npm and installed that way.

Without confidence that the issue is fixed, I can't merge. webpack-config-single-spa-ts has a dependency on fork-ts-checker-webpack-plugin, which has a peerDependency on typescript, so I'm not sure whether moving typescript to devDependencies is the right thing to do.

adripanico:
@joeldenning As far as I can see, `webpack-config-single-spa-ts` do have Typescript as peerDependency as well. So, at the end of the day, this change is just propagating the peerDepedency from `fork-ts-checker-webpack-plugin` to `webpack-config-single-spa-ts` and this one to the final app using it, right?

joeldenning:
The functionality of peerDependencies has changed so much and varies from package manager to package manager - I'm not sure what the correct approach is. Some have advocated for removing peerDependencies entirely because of how problematic they are.

Before merging, verifying that the change fixes the bug is required. I tried modifying webpack-config-single-spa-ts' package.json directly in node_modules and also yarn linking, and neither approach eliminated the error. Both of those approaches might not be equivalent to publishing to npm (linking notoriously is weird, and it's possible for package managers to reinstate the package.json during install), but I'm not sure how else to test it. Recommendations for how to verify the fix appreciated 🙏

Issue #382, opened 2023-06-10
Unable to load '@webpack-cli/serve' command


robmosca:
**Steps to reproduce the error:**

1. Create a new single-spa application (either root-config or application). Include Typescript
2. After the application has been created and the dependencies installed, run it with `pnpm start` or `npm start`

**Result:**

You get the error
```
[webpack-cli] Unable to load '@webpack-cli/serve' command
[webpack-cli] TypeError: options.forEach is not a function
    at WebpackCLI.makeCommand (/Users/roberto/Code/test-single-spa/root-config/node_modules/.pnpm/webpack-cli@4.8.0_webpack-dev-server@4.0.0_webpack@5.75.0/node_modules/webpack-cli/lib/webpack-cli.js:173:21)
    at ServeCommand.apply (/Users/roberto/Code/test-single-spa/root-config/node_modules/.pnpm/@webpack-cli+serve@1.7.0_webpack-cli@4.8.0_webpack-dev-server@4.0.0/node_modules/@webpack-cli/serve/lib/index.js:41:19)
    at loadCommandByName (/Users/roberto/Code/test-single-spa/root-config/node_modules/.pnpm/webpack-cli@4.8.0_webpack-dev-server@4.0.0_webpack@5.75.0/node_modules/webpack-cli/lib/webpack-cli.js:907:35)
    at async Command.<anonymous> (/Users/roberto/Code/test-single-spa/root-config/node_modules/.pnpm/webpack-cli@4.8.0_webpack-dev-server@4.0.0_webpack@5.75.0/node_modules/webpack-cli/lib/webpack-cli.js:1483:17)
    at async Promise.all (index 0)
    at async WebpackCLI.run (/Users/roberto/Code/test-single-spa/root-config/node_modules/.pnpm/webpack-cli@4.8.0_webpack-dev-server@4.0.0_webpack@5.75.0/node_modules/webpack-cli/lib/webpack-cli.js:1521:9)
    at async runCLI (/Users/roberto/Code/test-single-spa/root-config/node_modules/.pnpm/webpack-cli@4.8.0_webpack-dev-server@4.0.0_webpack@5.75.0/node_modules/webpack-cli/lib/bootstrap.js:11:9)
 ELIFECYCLE  Command failed with exit code 2.
```

**Proposed solution**
Updating the webpack-cli package to ^4.10.0 seems to prevent the problem.

PR #383, opened 2023-06-10
Update webpack-cli to 4.10

robmosca:
Fix for the problem described in issue https://github.com/single-spa/create-single-spa/issues/382

NOTE: When running the GitHub actions I had problems with the test step. The problem was that I was using a different version of pnpm on my local. I downgraded to the version previously used in the GitHub actions and could reproduce the problem. With that version of pnpm (6.29.1) my lock file was unstable and it changed when running tests. Also, the tests kept failing. By upgrading pnpm to the latest version (8.6.1) the tests run fine. Due to the update, I had to slightly adjust the commands in the GitHub workflow and add a dev dependency that was missing in the single-spa-welcome app. With these minor changes and adjustments, the GitHub actions workflow runs successfully in my local.

### DIFF
--- a/packages/single-spa-welcome/package.json
+++ b/packages/single-spa-welcome/package.json
@@ -30,6 +30,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.15.0",
+    "@babel/eslint-parser": "^7.15.0",
     "@babel/plugin-transform-runtime": "^7.15.0",
     "@babel/preset-env": "^7.15.0",
     "@babel/preset-react": "^7.14.5",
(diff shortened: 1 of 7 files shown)


TheMcMurder (commented):
Do we want to also make this change in the other packages?

TheMcMurder (approved):

packages/single-spa-welcome/package.json
```diff
@@ -30,6 +30,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.15.0",
+    "@babel/eslint-parser": "^7.15.0",
```
robmosca, line 33: This is required here: https://github.com/robmosca/create-single-spa/blob/9e99f853ae9ab6772450d5c14de7d027174daf26/packages/single-spa-welcome/.eslintrc#L3
I guess this was working before due to the different way the old version of pnpm was hoisting packages.

PR #393, opened 2023-12-06
fix: Bump various versions to remove vulnerable dependencies

ibacher:
I said I would work on updating the dependencies here. This is the first pass at it, designed to be very low-touch. Specifically the goal of this PR is resolve as many results reported from `pnpm audit` without moving to dependencies that are incompatible with Node 12, Webpack 4, etc., and this does that, with one exception, which results from `serve` depending on a version of `minimatch` that has a ReDoS issue in the generated configs.

In a near-future follow-on, I'll introduce changes that:

* Require Node 14 or 16 at a minimum
* Require Webpack 5


joeldenning (approved):
If the build passes, this looks good, thanks for your help. Please ping me on Slack in #core-team or #maintainers for the time being when there's a PR to review - my github queue is too large/stressful atm. I've added you as a maintainer to this repo so that your PRs automatically have the build run.

PR #402, opened 2024-04-08
infra: reconfigure jest

MilanKovacic:


joeldenning (commented):
Why has husky been removed?

joeldenning:
husky + pretty-quick is an alternative to lint-staged. I prefer pretty-quick because it doesn't require configuring all file extensions like lint-staged does (afaik)

This change removes husky without introducing lint-staged.

Perhaps we should change the scope of the PR to only include what is in the title - jest configuration - then a discussion and follow-up PR for husky or other changes?

.changeset/unlucky-suits-end.md
```diff
@@ -0,0 +1,5 @@
+---
```
ibacher, line 1: We should remove this .changeset.

ibacher (approved):
Aside from the changeset, LGTM

PR #406, opened 2024-07-06
fix types error in boilerplate

krisantuswanandi:
a

### DIFF
--- a/packages/generator-single-spa/src/root-config/templates/src/root-config.ejs
+++ b/packages/generator-single-spa/src/root-config/templates/src/root-config.ejs
@@ -11,8 +11,11 @@ registerApplication({
 
 // registerApplication({
 //   name: "@<%- orgName %>/navbar",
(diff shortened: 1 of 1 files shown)


joeldenning (approved):
Looks sensible, hopefully it matches the prettier expected indentations for both typescript and non-typescript, so that changes to these lines don't occur just when the user runs prettier on it.

joeldenning:
This pull request should have had a changeset. I don't know why the Github workflow didn't run, which would have caught the issue. I adjusted some Github settings so that hopefully future pull requests will have the workflow run. I'll try to retroactively add a changeset in preparation for a new release of create-single-spa.

joeldenning:
I undid the merge of this pull request and recreated it in https://github.com/single-spa/create-single-spa/pull/410, with a changeset.

joeldenning:
The equivalent changes in #410 were published in create-single-spa@4.1.6

PR #412, opened 2024-09-30
Default to output native modules. Add outputSystemJS option.

joeldenning:
This is a breaking change that changes all create-single-spa projects to default to outputting native ES modules.

### DIFF
--- a/.changeset/gentle-ducks-wonder.md
+++ b/.changeset/gentle-ducks-wonder.md
@@ -1,0 +1,5 @@
+---
+"webpack-config-single-spa": major
+---
+
+Output to native ES modules by default. Add new outputSystemJS option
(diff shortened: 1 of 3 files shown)


ibacher (approved):
As-is this seems correct. For Webpack, don't we also need to ensure `"type": "module"` is set in `package.json`?

joeldenning:
> For Webpack, don't we also need to ensure "type": "module" is set in package.json?

My understanding (might be incomplete) is that `"type": "module"` in the package.json is needed for webpack to know what kinds of source files it is dealing with, not the output format. I tested this change locally and webpack compiled without issue.

ibacher:
Ah! Perfect!

ibacher (approved):

joeldenning:
> Seems that my approvals are meaningless to GitHub, but that's fine with me.

I removed all write access, which might be why? I'm opening to changing a setting to make the approval checkmark green instead of gray, if it's possible without granting write access

PR #419, opened 2024-11-08
Full support for ESM in root configs. Resolves #418

### LINKED ISSUE #418: Create-single-spa root-config is broken.
Create-single-spa root-config is broken with latest changes. EJS file is still old template.EMS changes broken npm start.

Is there a way I could use old version to create single-spa root and child?

joeldenning:
See #418


packages/generator-single-spa/src/root-config/templates/src/layout/index.ejs
```diff
@@ -1,82 +0,0 @@
-<!DOCTYPE html>
```
joeldenning, line 1: Layout root configs used to contain the layout definition in the html, but that got moved to a separate file, so now two copies aren't needed.

PR #435, opened 2025-01-18
Add support for moduleFormat option for react/util mfes

joeldenning:


joeldenning:
Documented at https://github.com/single-spa/single-spa.js.org/pull/661

PR #440, opened 2025-02-04
Disable HMR in ES modules when outputSystemJS option is omitted

joeldenning:


packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -123,7 +125,7 @@ function webpackConfigSingleSpa(opts) {
         },
       },
       allowedHosts: "all",
-      hot: opts.outputSystemJS,
```
joeldenning, line 126: When this was undefined, it led to hmr being enabled.

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -39,6 +39,8 @@ function webpackConfigSingleSpa(opts) {
 
   let HtmlWebpackPlugin = opts.HtmlWebpackPlugin || _HtmlWebpackPlugin;
 
+  const outputSystemJS = !!opts.outputSystemJS;
```
joeldenning, line 42: Force a `true` or `false` rather than allowing undefined/null to be handled differently.

PR #446, opened 2025-02-25
restore webpack-dev-server index html

### LINKED ISSUE #445: "Your Microfrontend is not here" page missing
I believe #431 broke following screen:

Image

we're using `webpack-config-single-spa-react-ts`, with following config:
```
    outputSystemJS: true,
    orgName: "orgName",
    projectName: "projectName",
    webpackConfigEnv,
    argv,
  });
  
```

In versions "5.0.3", and below, running application without standalone mode, resulted in above screen. After updating to 6.0.0, a "Your Microfrontend is not here" page no longes appears, navigating to micro-frontend landing page results in blank screen, and `Uncaught ReferenceError: System is not defined` error in console.
Here's source of landing page
```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title>Webpack App</title>
  <meta name="viewport" content="width=device-width, initial-scale=1"><script defer src="[NAME].js"></script></head>
  <body>
  </body>
</html>
```

In #431 there was introduced variable
[`const isStandalone = webpackConfigEnv.standalone;`](https://github.com/single-spa/create-single-spa/blob/88364e898189e899050a6e3f90389367a39e9dcf/packages/webpack-config-single-spa/lib/webpack-config-single-spa.js#L38)
that's guarding usage of [`StandaloneSingleSpaPlugin`](https://github.com/single-spa/create-single-spa/blob/88364e898189e899050a6e3f90389367a39e9dcf/packages/webpack-config-single-spa/lib/webpack-config-single-spa.js#L143C1-L150C12)

In order to see that screen we need to have `StandaloneSingleSpaPlugin` [running with `disabled` flag passed](https://github.com/single-spa/standalone-single-spa-webpack-plugin/blob/70e9befc591bc1d49a4205b5ed744a9a4089bae3/lib/standalone-single-spa.js#L104C1-L115C6)

`webpack-config-single-spa` still passes this flag as `disabled: !webpackConfigEnv.standalone,`, which means we never can pass this flag

@joeldenning, your input would be highly apreciated

eckdanny:
Closes #445


packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -120,6 +121,15 @@ function webpackConfigSingleSpa(opts) {
         },
       },
       allowedHosts: "all",
+      hot: outputSystemJS,
```
joeldenning, line 124: The diff for this pull request looks wrong. The main branch already has this line, and also has `const outputSystemJS = !!opts.outputSystemJS`, but github's diff isn't showing that. Do you know what is causing the diff to be wrong? 

See related:
https://github.com/single-spa/create-single-spa/blob/88364e898189e899050a6e3f90389367a39e9dcf/packages/webpack-config-single-spa/lib/webpack-config-single-spa.js#L42

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -120,6 +121,15 @@ function webpackConfigSingleSpa(opts) {
...
+      setupMiddlewares: (middlewares, devServer) => {
+        const htmlDevServer = readFileSync(
+          path.join(__dirname, "index.html"),
+          "utf8"
+        );
+        devServer.app.get("/", (_, res) => res.send(htmlDevServer));
```
joeldenning, line 130: I like this implementation approach - avoid outputting the HTML file when building.

joeldenning (commented):
Please add a changeset, as per the CONTRIBUTING.md instructions.

Aside from the weird issue with Github diffing, this looks good.

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
eckdanny: prbly this: husky hook did something i didn't expect at `git add -p && git commit`, didn't notice till opening PR and then amended. I'll merge upstream and diff should be right

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
```diff
@@ -126,6 +127,14 @@ function webpackConfigSingleSpa(opts) {
       },
       allowedHosts: "all",
       hot: outputSystemJS,
+      setupMiddlewares: (middlewares, devServer) => {
+        const htmlDevServer = readFileSync(
+          path.join(__dirname, "index.html"),
+          "utf8"
+        );
+        devServer.app.get("/", (_, res) => res.send(htmlDevServer));
```
joeldenning, line 135:
```suggestion
        if (!opts.disableHtmlGeneration) {
          const htmlDevServer = readFileSync(
            path.join(__dirname, "index.html"),
            "utf8"
          );
          devServer.app.get("/", (_, res) => res.send(htmlDevServer));
        }
```

joeldenning (commented):
I tested this locally and it looks good. single-spa root configs use webpack-config-single-spa, but should not have this feature enabled, so I've added a suggestion.

packages/webpack-config-single-spa/lib/webpack-config-single-spa.js
joeldenning, line 135:
```suggestion
        if (!opts.disableHtmlGeneration) {
          const htmlDevServer = readFileSync(
            path.join(__dirname, "index.html"),
            "utf8"
          );
          devServer.app.get("/", (_, res) => res.send(htmlDevServer));
        }
```

joeldenning (approved):

joeldenning:
Released in https://github.com/single-spa/create-single-spa/releases/tag/webpack-config-single-spa%407.0.2

gregorysl:
I know this is bit late considering this PR is merged in, but this text says "Find the name **@emoney/home** and click on it.". Is there a way to have it dynamic?

eckdanny:
@gregorysl. 😮‍💨 that would be my fault

eckdanny:
opened #448

PR #448, opened 2025-02-28
fix(webpack-config-single-spa): fix copy in webpack-dev-server instructions

eckdanny:


joeldenning (approved):

joeldenning:
Released in https://github.com/single-spa/create-single-spa/releases/tag/webpack-config-single-spa%407.0.3 and related versions of projects that use webpack-config-single-spa

Issue #449, opened 2025-03-01
Unable to create Angular and Vue parcel apps using create-single-spa


armanossiloko:
Using version `5.0.7`, it appears to not be possible to create Angular and Vue apps.

I am honestly not sure what the reason behind this could be. Whether the CLI tool clashes with something else I have installed or not, I am not sure. There's no error shown or anything, it just exits after I type in my organization name.

Creating `svelte` and `react` apps is possible, however (though, the `react` one ends up with a bunch of errors which I believe I saw were reported as an issue already).

```
arman@DESKTOP-SR94ETS MINGW64 /d/Development
$ create-single-spa
? Directory for new project vue
? Select type to generate single-spa application / parcel
? Which framework do you want to use? vue
? Organization name (can use letters, numbers, dash or underscore) org

arman@DESKTOP-SR94ETS MINGW64 /d/Development
$
```

![Image](https://github.com/user-attachments/assets/ab941486-8b11-49c2-ba99-23b5d9307979)

![Image](https://github.com/user-attachments/assets/e004c677-abff-4c32-a304-48629198c60d)

armanossiloko:
After updating the package on my machine and retrying the Angular process, this is what I get:

![Image](https://github.com/user-attachments/assets/02669862-6572-4d88-a1d6-cb4156f5da6a)

As for Vue:

![Image](https://github.com/user-attachments/assets/dcd87edd-3cb8-43f0-9f90-4e2b2cfef545)

So the CLI tool fails while trying to execute `ng.cmd new angular-test` and `npx.cmd @vue/cli create vue-test --skipGetStarted`, but running both commands manually seems to work just fine. I honestly have no idea why this is failing when doing it via the `create-single-spa` command (perhaps it could be due to the command interactivity?).

joeldenning:
I'm working on single-spa-angular to support Angular 19, zoneless standalone components, and outputting native modules. Until that is working, I have no plans to fix create-single-spa's angular generator.

The vue generator works for me locally, and based on the error my guess is that it is platform-specific to Windows.

nickdeakin:
is this the fix you're after? https://github.com/single-spa/create-single-spa/issues/403#issuecomment-2662799618

armanossiloko:
> is this the fix you're after? [#403 (comment)](https://github.com/single-spa/create-single-spa/issues/403#issuecomment-2662799618)

This seems to have done the trick. Cheers.

Any reason why you haven't suggested this change via a Pull request?

PR #450, opened 2025-03-01
Added more console logs for failed Angular scaffolding #449

armanossiloko:
#449 - While I can't seem to find the root cause of the issue, I believe it's at least worthy of including a log in case the process for creating an Angular app just exits out of nowhere.

### DIFF
--- a/.changeset/witty-buses-warn.md
+++ b/.changeset/witty-buses-warn.md
@@ -1,0 +1,5 @@
+---
+"generator-single-spa": minor
+---
+
+Added console logging for errors that occur in Angular project creation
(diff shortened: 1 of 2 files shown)


joeldenning (changes requested):
Please add changeset and update the console logs, as described in https://github.com/single-spa/create-single-spa/pull/451#pullrequestreview-2652788037

armanossiloko:
> Please add changeset and update the console logs, as described in [#451 (review)](https://github.com/single-spa/create-single-spa/pull/451#pullrequestreview-2652788037)

I suppose it's been done. Not sure whether I did it incorrectly or not, but here it is.

joeldenning (approved):

PR #451, opened 2025-03-01
Added more console logs for failed Vue scaffolding #449

armanossiloko:
#449 - While I can't seem to find the root cause of the issue, I believe it's at least worthy of including a log in case the process for creating a Vue app just exits out of nowhere.

### DIFF
--- a/packages/generator-single-spa/src/vue/generator-single-spa-vue.js
+++ b/packages/generator-single-spa/src/vue/generator-single-spa-vue.js
@@ -79,7 +79,7 @@ module.exports = class SingleSpaVueGenerator extends Generator {
     );
 
     if (error) {
-      console.error(`The vue cli process to create a new project failed`);
+      console.error(`The Vue CLI process to create a new project failed.`);
       console.error(error);
 
       process.exit(1);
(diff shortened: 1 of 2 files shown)


packages/generator-single-spa/src/vue/generator-single-spa-vue.js
```diff
@@ -78,9 +78,21 @@ module.exports = class SingleSpaVueGenerator extends Generator {
       }
     );
 
+    if (error) {
+      if (error.message) {
+        console.error("The process encountered an error: ", error.message);
+      } else {
+        console.error("The process encountered an error: ", error);
+      }
```
joeldenning, line 86: Printing only `error.message` to the console will prevent the stacktrace or other information from being logged. Additionally, passing multiple arguments to console.error doesn't ensure a new line, which could make the output a bit less readable. The logged message also could be more clear about which process has failed

```suggestion
      console.error(`The vue cli process to create a new project failed`);
      console.error(error);
```

joeldenning (commented):
please add a changeset, per the CONTRIBUTING.md instructions

armanossiloko:
> please add a changeset, per the CONTRIBUTING.md instructions

I suppose it's been done. Not sure whether I did it incorrectly or not, but here it is.

joeldenning (approved):

